PR #1, opened 2020-09-30
Add command to toggle hidden files

unnamedd:
# Toggle hidden files
image

# How it works
Open Raycast and type "toggle hidden files".

# Result
On Finder, the user will be able to see hidden files and folder which starts with a dot.

# Screenshots

Difference between hidden files being presented or not

image

image


shpakovski:
Awesome, thanks 😍

PR #2, opened 2020-09-30
Add command to toggle dark mode

unnamedd:
# Toggle dark mode
toggle-dark-mode

# How it works
Open Raycast and type "toggle system appearance".

# Result
The system will alternate between the light mode and the dark mode.

### DIFF
--- a/system/toggle-dark-mode.applescript
+++ b/system/toggle-dark-mode.applescript
@@ -2,10 +2,10 @@
 
 # Required parameters:
 # @raycast.schemaVersion 1
-# @raycast.title Toggle Dark Mode
+# @raycast.title Toggle System Appearance
 # @raycast.mode silent
 
 # Optional parameters:
-# @raycast.icon 🦉
+# @raycast.icon 🌗
 
 tell application "System Events" to tell appearance preferences to set dark mode to not dark mode
--- a/system/toggle-dark-mode.applescript
+++ b/system/toggle-dark-mode.applescript
@@ -1,0 +1,11 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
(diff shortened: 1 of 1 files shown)


system/toggle-dark-mode.applescript
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Toggle Dark Mode
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.icon 🦉
```
PitNikola, line 9: Suggestion for emoji to reflect switching between dark/light: 🌗

system/toggle-dark-mode.applescript
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Toggle Dark Mode
```
PitNikola, line 5: How about naming it "Toggle System Appearance" since it's not only about the dark mode?

PitNikola (approved):
Nice one! We actually wanted to add it as built in command, but wanted to sort out permission request dialog in a nicer way first and it got de-prioritized. Let's have it as a script anyway, might be useful to some folks. I have only minor suggestion to the naming and emoji :)

system/toggle-dark-mode.applescript
unnamedd: I thought about this icon and I decided to use the owl anyway, but for sure, this one is better! Thanks for the suggestion! :D

system/toggle-dark-mode.applescript
unnamedd: Done!

PR #6, opened 2020-10-24
Create column-to-comma.sh

jdmcd:
Adds a script that uses the column data from your clipboard and converts it to comma separated. For example:

```
1
2
3
```

Becomes:

`1,2,3`


jdmcd:
@thomaspaulmann I just realized I created the file at the root level - let me update that

thomaspaulmann:
Great one!

How about creating a directory named `conversions` (or alternatively `format`) and add it there? Doesn't really fit the `system` one.

jdmcd:
No problem, updated!

thomaspaulmann:
Perfect and thanks for the quick update. Let's get it merged 🎉

PR #11, opened 2020-10-30
Added a Twitter section + new plugin

lazarnikolov94:
Added a Twitter Section and my Live Streaming Name Changer plugin within

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -16,6 +16,7 @@ This repository contains sample commands and documentation to write your own one
 - [Navigation](#navigation)
 - [Spotify](#spotify)
 - [System](#system)
+- [Twitter](#twitter)
 
 ## Bookmarks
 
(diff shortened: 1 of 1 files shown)


unnamedd:
@lazarnikolov94 you don't need to open a pull request changing the extensions.md or extensions.json because your change will be overwritten, both files are generated automatically via GitHub Actions. The best way for you to have your script documented on the documentation is open a pull request with your script command and defining these three metadatas:

```shell
# @raycast.author <Your Name>
# @raycast.authorURL <Your Social Media>
# @raycast.description <Description of your script command>
```

Check this example here https://github.com/raycast/script-commands/blob/master/system/toggle-system-appearance.applescript

If you need something else, please, let me know here or on Slack.

lazarnikolov94:
@unnamedd but my script is a JavaScript project, not a one-liner bash script. You can check it out on the link I added.

unnamedd:
> @unnamedd but my script is a JavaScript project, not a one-liner bash script. You can check it out on the link I added.

I read it when you sent on slack another day, but unfortunately your changes on those files will be overwritten by the bot as soon we have any new commit in this repository, even if your PR got approval and merge.

I've also created script command which was a single file, but it got to huge and I had transforme it into a Swift project and it's code won't be published here because we need to compile the project to run, then I suggest you to create a installer for your project (i.e: brew install live-streaming-name-changer) and then, create a script command just calling it (like you already have calling node index.js on/off). Of course, the user will have to install your lib on them machines, but in that way, will be easier to use.

I don't want to discourage you to submit your pull request, but Raycast still have some restrictions.

PR #13, opened 2020-10-30
System: Add script for starting screen saver

valentinchrt:
With the `Screen Saver` command, you will be able to start your current screen saver.

### DIFF
--- a/system/screensaver.sh
+++ b/system/screensaver.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Screen Saver
+# @raycast.mode silent
+
+# Optional parameters:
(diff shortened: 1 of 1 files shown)


system/screensaver.sh
```diff
@@ -0,0 +1,17 @@
...
+# @raycast.schemaVersion 1
+# @raycast.title Screen Saver
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.icon 🌀
+# @raycast.author Valentin Chrétien
+# @raycast.authorURL https://twitter.com/valentinchrt
+# @raycast.description A script command to start your current screen saver.
+# @raycast.packageName System
```
thomaspaulmann, line 13: You can drop this. It takes the parent directory as package name.

thomaspaulmann:
Thanks for your first contribution 🥳

PR #14, opened 2020-10-30
Update process of building and document generation

unnamedd:
The main goal of this pull request is the creation of the Makefile to make it easier the process of building and generation of documents using [Raycast Toolkit](https://github.com/raycast/script-commands/tree/master/Tools/Toolkit). Also, I updated the workflow instructions on documentation.yml.


thomaspaulmann:
Nice and simple, as always from you @unnamedd.

PR #16, opened 2020-10-31
Add Script Command: wayback-machine-save

crstauf:
Save clipboard URL to Wayback Machine.

Currently in the `developer-utils` directory, but would like a new, more appropriate directory; please advise, with consideration that there are a few other commands possible for the Wayback Machine and Internet Archive.


unnamedd:
> Save clipboard URL to Wayback Machine.
> 
> Currently in the `developer-utils` directory, but would like a new, more appropriate directory; please advise, with consideration that there are a few other commands possible for the Wayback Machine and Internet Archive.

You don’t need to change the folder, you can user the metatag `@raycast.packageName` to define a proper category for this script command. It will be part of the category `Developer Utils` but will live in a sub group with the name you decide.

crstauf:
@unnamedd I am asking which directory it should go in, as it's not only used by developers.

unnamedd:
Then I guess it will fit nicely on `bookmarks`, don’t you think?

crstauf:
I think that's probably the closest existing directory, though it doesn't function like a bookmark. How particular should we be about organizing?

unnamedd:
I guess we should be consistent to what the script command is doing, but I’m not sure where a script command for the wayback machine would fit, but I don’t think it’s a problem if you create a category where it will fit better, the only thing we should have is a folder where it is generically enough to fit more of these kind of scripts in the future.

crstauf:
Right; any thoughts on a generic directory?

Some possibles:
- `archive`
- `storage`

unnamedd:
Have a look to this [link](https://github.com/raycast/script-commands/blob/master/extensions.md#internet), I guess this is the best place where we can add it. Instead of `Internet Archive` we can let it as only `Internet` What do you think?

crstauf:
We can create a new directory, right?

unnamedd:
Yes, we can, but is it needed? If yes, do it, but we can use the benefit of the packageName! 😁

crstauf:
I've changed the package name to `Internet`, just need some input on a preferred directory name, as I don't think it fits in any of the other directories.

unnamedd:
Yeah buddy, I don’t have a strong opinion where your script command would fit better :/

developer-utils/wayback-machine-save.sh
```diff
@@ -0,0 +1,24 @@
...
+
+url=$(pbpaste)
+regex='(https?)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]'
+
+if [[ $url =~ $regex ]]
+then 
+	curl -s "http://web.archive.org/save/$url" > /dev/null
+	echo "Submitted $url to Wayback Machine"
+else
+	echo "$url (in clipboard) is not valid URL"
```
thomaspaulmann, line 23: Lets return here with a different exit code to signal that is an error, e.g. `exit(1)`

thomaspaulmann:
Let's keep the separate package name for now. As soon as we see more `Internet` related commands we can extract them to a separate directory. Thanks for this addition @crstauf. Please add the different exit code.

developer-utils/wayback-machine-save.sh
crstauf: @thomaspaulmann If I add `exit(1)`, then the error message doesn't show, so the user doesn't know what went wrong. Thoughts?

thomaspaulmann: Mmh... We should handle that case on our side. Maybe show something special and reopen Raycast...

crstauf: @thomaspaulmann Should I still add `exit( 1 )` at this point?

crstauf: Discussed in Slack; don't add `exit( 1 )` yet.

PR #17, opened 2020-10-31
Add Script Command: create-github-gist

crstauf:
Create a GitHub Gist from clipboard contents and copy Gist URL.


developer-utils/create-github-gist.sh
```diff
@@ -0,0 +1,48 @@
...
+
+##################################################################
+## Set GitHub username and personal access token and uncomment. ##
+##################################################################
+
+# GitHub username
+# user=
+
+# GitHub personal access token
+# access_key=
```
unnamedd, line 11: As GitHub is one of out most used services, I’d like to suggest here to, instead of every script command that might use GitHub credentials set manually the user and the access token in each file, why don’t we set those values just one and get it fromenvironment variables? 

What do you think?
We can define two env vars here: `GITHUB_ACCESS_TOKEN` and `GITHUB_USERNAME`, possibly, many of us already have the access token configured using env vars, then it will be better to be reused.

developer-utils/create-github-gist.sh
```diff
@@ -0,0 +1,48 @@
...
+# @raycast.packageName GitHub
+# @raycast.needsConfirmation true
+# @raycast.icon images/github-logo.png
+
+if [ -z ${user+x} ]; then
+	echo "GitHub username is missing.";
+	exit 1;
+fi
+
+if [ -z ${access_key+x} ]; then
```
unnamedd, line 31: You can make this check here and the one for the user requesting from the env var, like I suggested above.

unnamedd (commented):
Very cool script command! ;)

developer-utils/create-github-gist.sh
crstauf: @unnamedd I don't use environment variables myself, so I may need some assistance implementing both of your feedback items.

developer-utils/create-github-gist.sh
crstauf: @unnamedd I'm not clear what needs to change based on this feedback; can you please explain?

developer-utils/create-github-gist.sh
unnamedd: For sure, will be a pleasure to help you. Can we do this together tomorrow?

developer-utils/create-github-gist.sh
unnamedd: You won’t have to change much here, just check if the env var is empty or not! Nothing really special 😊

developer-utils/create-github-gist.sh
crstauf: Potentially, though I'm not sure precisely when. A little direction here may be enough for me to implement properly.

unnamedd: Sure.

First, you need to export those variable to your environment. To do this, open your `~/.bash_profile` using your preferred editor and write there:
```shell
export GITHUB_ACCESS_TOKEN=“your generated access token”
export GITHUB_USERNAME=“your-username”
```
Save it and back to your terminal, execute:
```shell
$ source ~/.bash_profile
```

developer-utils/create-github-gist.sh
unnamedd: Here, try this
```suggestion
if [ -z $GITHUB_ACCESS_TOKEN ]; then
```

developer-utils/create-github-gist.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Optional parameters:
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Create a GitHub Gist from clipboard contents and copy Gist URL.
+# @raycast.packageName GitHub
+# @raycast.needsConfirmation true
+# @raycast.icon images/github-logo.png
+
+if [ -z ${user+x} ]; then
```
unnamedd, line 26: and here, this

```suggestion
if [ -z $GITHUB_USERNAME ]; then
```

crstauf:
@unnamedd I tried implementing environment variables, but it seemed Raycast does not have access to environment variables...? If you put together a developed and tested version, I'll gladly implement.

unnamedd:
> @unnamedd I tried implementing environment variables, but it seemed Raycast does not have access to environment variables...? If you put together a developed and tested version, I'll gladly implement.

Hey @crstauf, sorry for the long delay, but I took a weekend off, tomorrow (monday) I will make some tests and call you on Slack, deal?

unnamedd (approved):

thomaspaulmann:
I love that one! So good and I thought about implementing it myself. Even better that you went ahead @crstauf. We should look into adding some higher level API to define API tokens for script commands that is nicely integrated. Also installing dependencies is something that I see more often. Wonder if there is a way to handle this more nicely...

PR #18, opened 2020-10-31
Fix: make screenshot command executable

crstauf:
Ran `chmod +x system/screensaver.sh`.


unnamedd:
I’m not sure if this change will have any effect on my machine, I guess in any case, I will have to set it as executable using chmod. What do you think?

crstauf:
@unnamedd I think all script commands in the git repo should be executable. There was talk about automating it I think somewhere in Slack.

unnamedd:
I completely agree with you, but I guess Raycast will be able to run it without have to set it as executable, I’m not sure now, I’m  not on computer, but tomorrow we can double check it if you want.

crstauf:
I was unable to find the command in Raycast prior to `chmod` adjustment.

unnamedd:
Huum... Weird... but, anyway, if worked for you, great! :D

crstauf:
I don't understand... will this be merged?

unnamedd:
Probably yes, but I don’t have permission to do this, only the Raycast developers. Sorry 😔

thomaspaulmann:
It would be great to automate this with a linter or bot that makes sure nothing non-executable is getting committed.

thomaspaulmann:
Thanks for fixing this @crstauf. Did you check if there are any other non-executable commands?

crstauf:
> Did you check if there are any other non-executable commands?

@thomaspaulmann I did not; I tried to use the screensaver command and discovered it wasn't indexed. Perhaps a Raycast Script Command to check for any Raycast Script Commands that are not executable? ;)

thomaspaulmann:
Haha inception :D

unnamedd:
> It would be great to automate this with a linter or bot that makes sure nothing non-executable is getting committed.

I will check that!

thomaspaulmann:
Yes, the `FileManager` exposes a [method](https://developer.apple.com/documentation/foundation/filemanager/1414159-isexecutablefile) to check if a file is executable.

unnamedd:
Good to know @thomaspaulmann. For the Toolkit I'm using TSC (Tool Supporting Core) to read the files, as it is written on top of FileManager, I'm sure they added support for this kind of check there!

PR #19, opened 2020-10-31
Completion messages for scripts in dev utils added

japanese-goblinn:


crstauf:
@japanese-goblinn Doesn't the script mode needs to be changed to `compact` in order for the completion message to be displayed?

japanese-goblinn:
> @japanese-goblinn Doesn't the script mode needs to be changed to `compact` in order for the completion message to be displayed?

I guess no, completion message is showing up. In `compact` mode `Raycast` window will not closed and in my opinion this is not an appropriate behavior for this kind of scripts.

Screen Shot 2020-10-31 at 18 42 28

crstauf:
> I guess no, completion message is showing up.

Oh cool!!! 😎 Didn't know that was possible.

developer-utils/decode-base64.sh
```diff
@@ -8,4 +8,5 @@
 # Optional parameters:
 # @raycast.icon 💻
 
-pbpaste | base64 -d | pbcopy
\ No newline at end of file
+pbpaste | base64 -d | pbcopy
+echo "Decoded"
```
unnamedd, line 12: To make this echo be visible, you need to change the metatag `@raycast.mode` to `compact` instead of the current `silent`. 

This comment applies for the next two files (`encode-base64.sh` and `generate-uuid.sh`).

unnamedd (changes requested):
Changes on the metadata is needed, to make these changes effective.

japanese-goblinn:
Are you sure about this? Don't you think that this change will break script workflow, because now user need to manually close window to paste the result?

unnamedd:
The only reason for me to suggest that is I used to expect the result on the same place. The HUD is amazing, for sure, but in my case, I use three huge monitors, so won’t be easy for me to see that HUD. At least, I will see result on the result box and if I decide to do anything else, Raycast is already opened.

thomaspaulmann:
I think we should keep the `silent` mode for these utility scripts. The `HUD` completion message is a neat addition to give the user the verification that the command was successful. `Compact` mode wouldn't be the right approach here because most likely you don't continue in Raycast after a conversion and rather go to some other tool, e.g. the terminal to paste in your decoded content. Gonna merge it and thanks to @japanese-goblinn for your first contribution. Super appreciated!

PR #24, opened 2020-11-03
Add Script Commands: Hex to RGB(A)

crstauf:


conversions/hex-to-rgb.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Hex to RGB
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Convert hex to RGB.
```
thomaspaulmann, line 11: I think it's worth to mention that this reads from the clipboard.
```suggestion
# @raycast.description Convert HEX color values in your Clipboard to RGB values.
```

conversions/hex-to-rgba.sh
```diff
@@ -0,0 +1,33 @@
...
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Hex to RGBA
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Convert hex to RGBA.
```
thomaspaulmann, line 11: Same here.

thomaspaulmann (approved):
Nice one! That's a neat addition.

PR #26, opened 2020-11-03
Adjust permissions to be executable

crstauf:


thomaspaulmann:
Thanks for going over the existing commands and make them executable. Curious, what command did you use to identify the non-executable ones? This might help to automate it.

crstauf:
> what command did you use to identify the non-executable ones?

@thomaspaulmann I actually used `find . -perm +111 -type f` to find the executable ones, and then went through the file list cross-referencing, and if I didn't find it ran `chmod +x`.

PR #30, opened 2020-11-06
Add Script Command: current-weather

crstauf:


unnamedd (approved):
Great addition @crstauf! 😍

thomaspaulmann:
Ha, you beat me on this command! Looks like `https://wttr.in/?format="%c+%f+%w+in+%l` spits out a rather long string. Should we cap it somehow with an regex?

crstauf:
@thomaspaulmann It fits for me. I suspect that the `+in+%l` is the problem, as that's the location, and that could be very long for some and very short for others. I decided to include the location in case the geolocation was wrong, so the user had the opportunity to confirm.

crstauf:
@thomaspaulmann Limited the location output to everything before the first comma in fad4c71.

thomaspaulmann:
Nice! This makes it more streamlined with our UI. Gonna get that straight into my favorites :-)

PR #33, opened 2020-11-07
Add Script Commands: Music

crstauf:


unnamedd (changes requested):
There's a folder called `spotify`, I suggest to rename that folder to `music` and add `packageName` to those script commands to `Spotify` and add your script commands on that "new" folder . 

In the beginning they only developed Spotify controls, but now, with more controls but for Music, doesn't make any sense for me to have an special folder for Spotify.

system/music-next.sh
```diff
@@ -0,0 +1,19 @@
...
+# @raycast.schemaVersion 1
+# @raycast.title Next
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Next track in Music
+# @raycast.packageName Music
+# @raycast.icon images/music-logo.png
```
unnamedd, line 13: Where's this file?

crstauf: Ahh my mistake, didn't pull it over from local directory into git repository.

crstauf:
> I suggest to rename [`spotify`] folder to `music`

Will wait for additional thoughts on file structure.

thomaspaulmann:
@unnamedd I agree with you that we we should rename the directory from `Spotify` to `Music`. It was a poorly chosen name from me. Question, how do we deal with folks that already use the `Spotify` controls? I assume they would need to re-add the music folder after pulling.

thomaspaulmann (changes requested):
Let's use `applescript` as file extension to keep it aligned with the other Applescripts we have. Did you simply copy the template? Do we need another template for Applescript?

unnamedd:
> @unnamedd I agree with you that we we should rename the directory from `Spotify` to `Music`. It was a poorly chosen name from me. Question, how do we deal with folks that already use the `Spotify` controls? I assume they would need to re-add the music folder after pulling.

Unfortunately yes, but it’s a good trade of, don’t you think?

thomaspaulmann:
> Unfortunately yes, but it’s a good trade off, don’t you think?

Generally yes, but I don't like breaking changes. In this case it was my fault and the lesson learned is to name our directories generic enough to make sure we don't need to rename them. 

We can also have subdirectories to make sure that related commands live close together on disk. I'm leaning to introduce directly some sub-directories, f.e. Spotify, to avoid another breaking change later on.

unnamedd:
> We can also have subdirectories to make sure that related commands live close together on disk. I'm leaning to introduce directly some sub-directories, f.e. Spotify, to avoid another breaking change later on.

I don't really think we need to have subdirectories. We already have the `packageName`, what in my opinion is enough to gives us the difference we need. More subdirectories, more complexity to search for the scripts, more time spent.

crstauf:
> Let's use `applescript` as file extension to keep it aligned with the other Applescripts we have. Did you simply copy the template? Do we need another template for Applescript?

I noticed that the `screen saver` command was Applescript and has a `sh` extension, so figured that was okay. That was fixed in the repository at some point, but in the directory that Raycast actually uses it was an `sh`. Additionally my editor doesn't syntax highlight `applescript` files as well as it does `sh`.

Now that I know `applescript` extension needs to be used I'll do that for now on.

shpakovski:
Looks great 👍 Thanks

PR #34, opened 2020-11-07
Add Script Command: wallpaper-refresh

crstauf:


thomaspaulmann (changes requested):
Great addition! I didn't know this is possible 👏 

Same as for the other Applescripts, lets keep the file extension consistent and change it from `.sh` to `.applescript`

shpakovski:
👍

PR #35, opened 2020-11-07
Add Script Commands: TV

crstauf:


system/tv-show.sh
```diff
@@ -0,0 +1,17 @@
...
+# Optional parameters:
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Show TV
+# @raycast.packageName TV
+# @raycast.icon images/tv-logo.png
+
+tell application "TV"
+	activate
+end tell
```
thomaspaulmann, line 17: What's the difference of this command to just launch the `TV` app from Raycast?

thomaspaulmann (changes requested):
You are on fire 🔥 Thanks for this addition. Can we get a toggle (play/pause) command as well? I find this very useful because you can search for the same term. As to the `Show TV` command, is this really needed? Seems like it doesn't do something different to just launch the `TV` app, right?

system/tv-show.sh
crstauf: Oh, fair enough; will remove the `show` command.

My intent was to focus the window, but using the app launcher works the same.

crstauf:
> can we get a toggle command

@thomaspaulmann I had started with a toggle, and then decided I preferred separate commands, as it's clearer what the expected result is, and having two commands that do the same (within certain context) seems strange. Thoughts?

PR #38, opened 2020-11-08
Useful Safari scripts

japanese-goblinn:


shpakovski:
Fantastic, thanks!

crstauf:
Based on previous conversation on https://github.com/raycast/script-commands/pull/33#issuecomment-723560063, directory name of `safari` should be changed to something more generic, perhaps `browsing`?

PR #39, opened 2020-11-08
Scripts to open Finder directory in Terminal and vice versa

japanese-goblinn:


navigation/open-finder-from-terminal.applescript
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Terminal -> Finder
```
shpakovski, line 5: Could you please rename to smth like “Open current iTerm directory in Finder”?

navigation/open-terminal-from-finder.applescript
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Finder -> Terminal
```
shpakovski, line 5: Same here, but vice versa 🙏

shpakovski (changes requested):
Thanks!

navigation/open-finder-from-terminal.applescript
japanese-goblinn: yep

PR #41, opened 2020-11-08
Change 'safari' directory name to 'browsing'

crstauf:
Based on previous conversation on https://github.com/raycast/script-commands/pull/33#issuecomment-723560063, change `safari` directory (intro'd in #38) to more generic name `browsing`.


japanese-goblinn:
Yep, a good point. Also folks here should not forget to add some kind of prefix (I suggest `safari_`) to scripts name to show clearly that they're works only with `Safari`

shpakovski:
Perfect, thanks!

PR #42, opened 2020-11-08
Enhance IP address commands to use inline mode

crstauf:
Because of the _Copy Output_ action (`CMD+.`) enhance the IP address commands to use `inline` mode.

### DIFF
--- a/developer-utils/get-external-ip-v4.sh
+++ b/developer-utils/get-external-ip-v4.sh
@@ -3,13 +3,12 @@
 # Required parameters:
 # @raycast.schemaVersion 1
 # @raycast.title External IPv4
-# @raycast.mode compact
+# @raycast.mode inline
+# @raycast.refreshTime 1h
 
 # Optional parameters:
 # @raycast.icon 🌐
 # @raycast.packageName Internet
 
 ip=$(curl -4 -s -m 5 https://ifconfig.co)
-echo $ip | pbcopy
-
-echo "Copied $ip"
+echo $ip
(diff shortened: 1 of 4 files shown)


unnamedd:
I didn't understand the reason why the `pbcopy` was removed from the scripts, one of the great benefit of those scripts was exactly the copy of the IP v4/v6 to the clipboard. The `inline`+`refreshTime` is great, but in those specific cases, I'd rather copy of the information instead of the visualization.

crstauf:
> I didn't understand the reason why the `pbcopy` was removed from the scripts

@unnamedd The _Copy Output_ action makes the `pbcopy` unnecessary.

> the great benefit of those scripts was exactly the copy of the IP v4/v6 to the clipboard

The _Copy Output_ action copies the output to the clipboard. Changing the command `inline` gives all the current functionality, with the added benefit of a quick glance to check the IP address rather than pasting somewhere.

unnamedd:
I download two of the modified scripts here ([this](developer-utils/get-external-ip-v4.sh) and [this](https://github.com/raycast/script-commands/blob/211bd5068696775f5beb04b1fbdb11d7144e7b68/developer-utils/get-local-ip-v4.sh) and I didn't achieve the expected result. Did I miss something?

crstauf:
@unnamedd These steps should work:

1. Open Raycast
2. Select Local IPv4 inline Script Command
3. Press `ENTER` to refresh your IP address
4. Press `CMD + .` to copy the command's output
5. Paste clipboard somewhere to confirm

Here's a video of me executing the above steps: https://cloudup.com/c3CTZ8kMTw4.

unnamedd:
Indeed @crstauf, make sense. I forgot completely about the `Copy Output` shortcut! 

I agree with you the use of `pbcopy` is not needed, but I'm afraid others might forget the `Copy Output` as well, like me.
Make sense to wait for more opinions.

shpakovski:
🤯

PR #43, opened 2020-11-08
Add detailed output for GitHub Unread Notifications command

crstauf:
While output of total number of notifications is fine, we can do better: output total number per [notification reason](https://docs.github.com/en/free-pro-team@latest/rest/reference/activity#notification-reasons).


shpakovski:
Nice one 👍

PR #46, opened 2020-11-09
Add Script Commands: ping

crstauf:
`ping` and `ping-monitor`


unnamedd (approved):

developer-utils/ping.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# @raycast.schemaVersion 1
+# @raycast.title Ping address in clipboard
```
PitNikola, line 4: Let's name file `ping-address-from-clipboard`

developer-utils/ping.sh
```diff
@@ -0,0 +1,14 @@
...
+# @raycast.mode fullOutput
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Ping an IP address or URL in clipboard.
+# @raycast.packageName Internet
+# @raycast.needsConfirmation true
```
PitNikola, line 10: Do you think this command worth having a confirmation? Doesn't seem to be that dangerous...

developer-utils/ping.sh
```diff
@@ -0,0 +1,14 @@
...
+# @raycast.icon 🌐
+
+clipboard=$(pbpaste)
+ping -i 0.25 -t 3 "$clipboard"
```
PitNikola, line 14: How about using url validation here? See https://github.com/raycast/script-commands/blob/master/navigation/open-url-from-clipboard.sh

PitNikola (changes requested):

developer-utils/ping.sh
crstauf: Anytime I use the clipboard for specific content, I include a request for confirmation, as a double-check that clipboard contains valid content.

developer-utils/ping.sh
crstauf: `ping` is not limited to URLs, can also be IP addresses.

developer-utils/ping.sh
crstauf: Changed to `ping-from-clipboard.sh` to be inclusive of both URLs and IP addresses.

PitNikola (approved):

Issue #47, opened 2020-11-10
Proposal: make packageName required meta


crstauf:
If `packageName` was a required meta key, that allows for (at least) a couple things:

- increased portability of the script command (a user can cherry pick scripts outside of directories and not lose any context info)
- turn off entire packages rather than (or in addition to) individual scripts

While portability is important, I see the ability to turn off entire packages to be a critical step. For example, I use Cloudup as my screenshot sharing service, but there are plenty of other services that do that. A user of another service won't need Cloudup, and so would need to search the entire list for Cloudup scripts, or delete the files from their directory each time they update the scripts from this repository. If Raycast supported toggling entire packages, that'd be a one time change, and done.

I expect there are additional advantages to requiring `packageName`, whether now or in the future.

maximkrouk:
Also, a package browser might be a nice feature if the proposal will be accepted. Maybe it's even worth making a fully package-based script import system instead of a folder-based (like in Xcode SPM with FolderURLs support).

PitNikola:
Thanks for putting together this proposal @crstauf!

We had a discussion about this change and decided to not make `packageName` a required field. We actually had it required in our initial implementation, but then decided to make it less restrictive because in the end it's just a label in the list. It doesn't represent an ID, you can have script commands from different directories with same `packageName`, you can use it as a category and so on. Perhaps naming could've been better.

However, portability is indeed a good point, so we decided to make it a requirement to provide `packageName` for all scripts in this repository. I will add this section to the guidelines.

As to package based systems, it's deliberate decision to keep script commands as small, simple and self contained commands, so we want to avoid introducing complexity with any sort of package management. For more complex extensions there will be API.

Please let me know if you have any questions.

PR #49, opened 2020-11-10
Moved media scripts in separate directories

japanese-goblinn:
For me this solves problem that I'm not using `Apple Music` and don't want it in Raycast suggestions, so I can only add `media/spotify` and that's all. If you want both - just add two folders.


unnamedd:
Today, Raycast or the [Toolkit](https://github.com/raycast/script-commands/tree/master/Tools/Toolkit) doesn't reads the content of subdirectories, whether to add the scripts on the Script Commands Preferences in Raycast nor in the documentation files ([json](https://github.com/raycast/script-commands/blob/master/extensions.json) and [markdown](https://github.com/raycast/script-commands/blob/master/extensions.md)). 

That being said, I'd like to ask: why not disable the script commands you're not using? 
Giving me as an example: I don't use Spotify and then, all those script commands are disabled.

crstauf:
I submitted a proposal that would support resolving this: #47.

japanese-goblinn:
> That being said, I'd like to ask: why not disable the script commands you're not using?

because I don't what do disable every unused script manually

> Giving me as an example: I don't use Spotify and then, all those script commands are disabled.

If you want to use only Apple Music just add `media/apple-music`, if you want Spotify - just add `media/spotify`, if you want both - just add both 😅. I think its better to add a subdirectory that contains scripts related to music service than deactivate manually every script that I'm not using.

Screen Shot 2020-11-10 at 20 15 00

crstauf:
This was discussed previously in https://github.com/raycast/script-commands/pull/33#issuecomment-723565321, and determined that the use of `packageName` is preferable.

Please see https://github.com/raycast/script-commands/pull/49#issuecomment-724840882.

japanese-goblinn:
> That being said, I'd like to ask: why not disable the script commands you're not using?

because I don't what do disable every apple script

> Giving me as an example: I don't use Spotify and then, all those script commands are disabled.

I don't understand. If you want to use only Apple Music just add `media/apple-music`, if you want Spotify - just add `media/spotify`, if you want both - just add both 😅. I think its better to add a subdirectory that contains scripts related to music service than deactivate manually every commit that I'm not using.

Screen Shot 2020-11-10 at 20 15 00


> This was discussed previously in [#33 (comment)](https://github.com/raycast/script-commands/pull/33#issuecomment-723565321), and determined that the use of `packageName` is preferable.
> 
> Please see [#49 (comment)](https://github.com/raycast/script-commands/pull/49#issuecomment-724840882).

Sorry, I didn't see. No questions now.

unnamedd:
> If you want to use only Apple Music just add `media/apple-music`, if you want Spotify - just add `media/spotify`, if you want both - just add both 😅. I think its better to add a subdirectory that contains scripts related to music service than deactivate manually every script that I'm not using.

I continue to disagree with you that we need to have subdirectories, but you have a point, and what maybe @thomaspaulmann , @PitNikola and @shpakovski can discuss it there, about an implementation of sub categories like we have in the [documentation](https://github.com/raycast/script-commands/blob/master/extensions.md) based on the `packageName` and making them possible to enable/disable the whole subcategory.

In my opinion, this should be enough without the need to create subdirectories creating more complexity to everyone.

japanese-goblinn:
> > If you want to use only Apple Music just add `media/apple-music`, if you want Spotify - just add `media/spotify`, if you want both - just add both 😅. I think its better to add a subdirectory that contains scripts related to music service than deactivate manually every script that I'm not using.
> 
> I continue to disagree with you that we need to have subdirectories, but you have a point, and what maybe @thomaspaulmann , @PitNikola and @shpakovski can discuss it there, about an implementation of sub categories like we have in the [documentation](https://github.com/raycast/script-commands/blob/master/extensions.md) based on the `packageName` and making them possible to enable/disable the whole subcategory.
> 
> In my opinion, this should be enough without the need to create subdirectories creating more complexity to everyone.

Yep, absolutely agree.

maximkrouk:
Hi there. +1 for packages I already gave a little feedback [here](https://github.com/raycast/script-commands/issues/47#issuecomment-724859199), but also +1 for manually adding subdirectories as script directories, instead of disabling scripts from shared directories, it's much more inconvenient.

PR #50, opened 2020-11-10
Add Script Command: set-scripts-executable

crstauf:


crstauf:
> very specific to this repository

@PitNikola How do you mean? The `directories` variable allows any directories to be set, and the list of file extensions is only what I'm familiar with because of Raycast. Anyone could add additional file extensions, and submit a PR for those as well.

crstauf:
> want to resolve this problem on Raycast side

Should I close this then?

PitNikola:
I mean it has references of Toolkit and have particular list of extensions. It also requires specification of directories in the file itself making it less reusable / portable. Let's close this one and I've added it to the list of improvements that we need to do.

PR #56, opened 2020-11-12
Add Script Commands: Convert epoch to human readable date

kastnerorz:
Example: Convert 1605146036 to 2020-11-12 09:53:56

### DIFF
--- a/conversions/epoch-to-human-date.sh
+++ b/conversions/epoch-to-human-date.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Convert Epoch to Human-Readable Date
+# @raycast.mode silent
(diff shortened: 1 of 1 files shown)


conversions/epoch-to-human-date.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Convert Epoch to Human-Readable Date
+# @raycast.mode silent
+#
+# Optional parameters:
+# @raycast.icon ⏱
+# @raycast.currentDirectoryPath ~
```
PitNikola, line 10: Let's remove this one as it's not needed for this particular script.

kastnerorz: I fixed it. 🎉

PR #58, opened 2020-11-12
fix(Finder & Terminal): iTerm support removed

maximkrouk:
The script fails if iTerm is not installed. Now script will use `Terminal.app` only as a quick fix. iTerm support will be requested in a separate PR 🙂


japanese-goblinn:
After this PR will be merged I'll create a new one with support of `iTerm` so everyone can decide which one to use. Me and @maximkrouk haven't figure out how to check if app is installed in `AppleScript` so with few trade offs this is a good solution for this problem.

PR #61, opened 2020-11-12
Increase refreshTime of current-weather

crstauf:


thomaspaulmann (approved):
Thanks for acting quickly on this front and leaving a comments for other contributors. 🌤️

PR #62, opened 2020-11-12
Add Script Command: Generate Password

hofmannsven:
This script generates a random password with 40 chars and copies it to the clipboard.

Example: LsK9XN5CYq4YHFTR4gADCfne/Jtn/I+AbHoUeLJo

PR #63, opened 2020-11-12
Add Script Command: xkcdpass

crstauf:
https://pypi.org/project/xkcdpass/


PitNikola (changes requested):
As discussed in Slack and per [new section in guidelines about dependencies](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#scripts-that-require-installation-of-dependencies), let's add a comment section explicitly stating the dependency and installation instruction. Thanks!

crstauf:

commands/communication/xkcdpass.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Dependency: requires xkcdpass (https://pypi.org/project/xkcdpass/).
+# Install via pip: `pip install xkcdpass`
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title xkcdpass
```
thomaspaulmann, line 8: Let's give it a proper title.
```suggestion
# @raycast.title Generate Passphrase
```

thomaspaulmann (changes requested):
Nice one. What's the difference to #62? Do we need both?

crstauf:
> What's the difference to #62? Do we need both?

Yes.

> There are dozens of random password generators out there that will happily put together a bunch of random characters for you to use as a password. These random passwords are secure, but they're a huge pain to actually remember.
> 
> Random passphrases provide the best combination of memorability and security.

Source: [useapassphrase.com](https://www.useapassphrase.com/#i-get-it-simple-passwords-are-cracked-easily-but-why-should-i-use-a-random-pass-phrase-instead-of-say-ipz2-az8k-0h-)

PR #64, opened 2020-11-13
Add Script Commands: DND

crstauf:


PitNikola (changes requested):
As discussed in Slack and per [new section in guidelines about dependencies](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#scripts-that-require-installation-of-dependencies), let's add a comment section explicitly stating the dependency and installation instruction.

crstauf:

commands/system/dnd-off.sh
```diff
@@ -0,0 +1,35 @@
...
+# Dependency: do-not-disturb-cli (https://github.com/sindresorhus/do-not-disturb-cli)
+# Install via npm: `npm install --global do-not-disturb-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Do Not Disturb - Off
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Uses sindresorhus/do-not-disturb-cli to turn on DND.
+# @raycast.icon 🌗
```
thomaspaulmann, line 12: We ave the moon for toggle appearance already. 

```suggestion
# @raycast.icon 😴
```

commands/system/dnd-on.sh
```diff
@@ -0,0 +1,35 @@
...
+# Dependency: do-not-disturb-cli (https://github.com/sindresorhus/do-not-disturb-cli)
+# Install via npm: `npm install --global do-not-disturb-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Do Not Disturb - On
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Uses sindresorhus/do-not-disturb-cli to turn on DND.
+# @raycast.icon 🌗
```
thomaspaulmann, line 12: Same as above.

commands/system/dnd-on.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# Dependency: do-not-disturb-cli (https://github.com/sindresorhus/do-not-disturb-cli)
+# Install via npm: `npm install --global do-not-disturb-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Do Not Disturb - On
```
thomaspaulmann, line 7:
```suggestion
# @raycast.title Turn On Do Not Disturb
```

commands/system/dnd-off.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# Dependency: do-not-disturb-cli (https://github.com/sindresorhus/do-not-disturb-cli)
+# Install via npm: `npm install --global do-not-disturb-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Do Not Disturb - Off
```
thomaspaulmann, line 7:
```suggestion
# @raycast.title Turn Off Do Not Disturb
```

thomaspaulmann (changes requested):
A great one! I can use that quite often myself :-) 

Generally I would use a single command and toggling the do not disturb mode. There isn't much value having different command for turning it on/off. 

I was also digging into the [dependency tree](https://github.com/sindresorhus/do-not-disturb), at seems that you could build this entire command with `Swift`, maybe even with `AppleScript` and could avoid installing dependencies. This would make it easier for others to adopt the command. This is probably out of scope for this PR but I thought to bring it up.

crstauf:
> There isn't much value having different command for turning it on/off

Unless you're not sure what the current state is, and want to make sure you've got the correct mode active.

crstauf:
> build this entire command with `Swift`, maybe even with `AppleScript`

I'd have no problem with this command being replaced by a native Swift or Applescript version.

thomaspaulmann:
> > build this entire command with `Swift`, maybe even with `AppleScript`
> 
> I'd have no problem with this command being replaced by a native Swift or Applescript version.

We should rather strive for the solution that makes most users happy. Having to install less dependencies is a more flawless experience. I don't think that people would just re-implement existing commands when there is one already committed.

PR #65, opened 2020-11-13
Add Script Commands: emoji

crstauf:


PitNikola (changes requested):
As discussed in Slack and per [new section in guidelines about dependencies](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#scripts-that-require-installation-of-dependencies), let's add a comment section explicitly stating the dependency and installation instruction.

crstauf:

commands/communication/emojis/emoji-copy.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Dependency: emoj (https://github.com/sindresorhus/emoj)
+# Install via npm: `npm install --global emoj`
+
+# @raycast.schemaVersion 1
+# @raycast.title Copy emoji
```
thomaspaulmann, line 7: Not sure if `Copy emoji` is a great name for this command. I couldn't tell what it does. How about this title?

```suggestion
# @raycast.title Search Emoji From Clipboard
```

commands/communication/emojis/emoji-copy.sh
```diff
@@ -0,0 +1,30 @@
...
+# @raycast.mode compact
```
thomaspaulmann, line 8: 👍

commands/communication/emojis/emojis-search.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Dependency: emoj (https://github.com/sindresorhus/emoj)
+# Install via npm: `npm install --global emoj`
+
+# @raycast.schemaVersion 1
+# @raycast.title Get emojis
```
thomaspaulmann, line 7:
```suggestion
# @raycast.title Search Related Emojis from Clipboard
```

commands/communication/emojis/emojis-search.sh
```diff
@@ -0,0 +1,33 @@
...
+# @raycast.mode fullOutput
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Search emojis for clipboard value.
+# @raycast.needsConfirmation true
```
thomaspaulmann, line 12: Lets remove the `needsConfirmation`.

commands/communication/emojis/emoji-copy.sh
```diff
@@ -0,0 +1,30 @@
...
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Copy first emoji found from clipboard value.
+# @raycast.packageName Communication
+# @raycast.needsConfirmation true
```
thomaspaulmann, line 13: I don't think you need a confirmation here.

thomaspaulmann (changes requested):
Lets give the commands somewhat more descriptive titles. Otherwise they appear very general and are not easy to understand what they do.

commands/communication/emojis/emoji-copy.sh
crstauf: This command searches for an emoji based on text in clipboard, and copies the first result to the clipboard. I'd like to give some indication to the user that the first result is copied.

commands/communication/emojis/emojis-search.sh
crstauf: As mentioned elsewhere, to me it's preferable to do a check to the user before the script accesses their clipboard, to confirm that they've the correct content there.

commands/communication/emojis/emoji-copy.sh
crstauf: As mentioned elsewhere, to me it's preferable to do a check to the user before the script accesses their clipboard, to confirm that they've the correct content there.

thomaspaulmann: I'm not really convinced that this is a big issue. All commands are transparent and when you install them should be treated as safe. The confirmation dialog should be only used for irreversible actions, e.g. deleting files. I'll accept that one but I would please you to remove it from almost all commands to fit into the user experience of the rest of Raycast.

thomaspaulmann (approved):

PR #67, opened 2020-11-13
Add Script Command: is-online

crstauf:


PitNikola (changes requested):
As in previous PRs let's add a comment section explicitly stating the dependency and installation instruction.

crstauf:

fe9lix:
@crstauf What would be the concrete use case for this script? As far as I understand, it monitors the internet connection, but then how would it be useful within Raycast? A refresh timer is not "instant" enough and we don't yet have notifications. And as one-off script to confirm that my connection is currently off, I also can't really see the benefit since I would usually follow the information that OS utilities give me here.

Also, we have a bit of a similar question as with the is-up script, i.e. whether this would justify an explicit installation and dependency.

crstauf:
> concrete use case for this script

Accessing Raycast script command is much faster than accessing system level utilities, or browsing to `google.com`.

> a refresh timer is not "instant" enough

So you run it when you suspect your Internet connection is out to confirm.

PR #68, opened 2020-11-13
Add Script Command: pageres

crstauf:
I did have to [do a little weirdness](https://github.com/raycast/script-commands/compare/master...crstauf:add/pageres?expand=1#diff-06a50adaf152a72417c009f4e4d1454a3c98617760ece931cfae51fd867e01f5R16) in order to get the command to work.

[Posted in Slack](https://raycastcommunity.slack.com/archives/C01D1B36344/p1605247189239000) requesting help:

> due to permissions troubles, i was forced to create a new global directory for `npm` packages… while a command works from my Terminal, i can’t get the same command to work from a Script Command… i’ve tried everything i know but not working. can someone with more knowledge on this please advise?


crstauf:

browsing/website-screenshots.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash -l
```
PitNikola, line 1: Did you figure out why exactly it requires login shell? Usually it means that there is some dependency that you load from your profile so it's not guaranteed to work on others people machines. As a rule of thumb, I would avoid having `#!/bin/bash -l` in this repo's script commands as much as possible for better portability.

P.S. Now I noticed that your google scripts also have `-l` in shebang. Do they stop working without it?

PitNikola (changes requested):

crstauf:

browsing/website-screenshots.sh
crstauf: Why would it not work on other systems? The `-l` was required for it to work. @unnamedd tested and it worked for him with the `-l`.

PitNikola: `-l` means using login shell which means the process running the script will loads all your profiles (such as `.bash_profile`), set up your environmental variables and exports $PATH. This means that if it works for you only with '-l', there is a high chance it will not work for somebody else. Plus it will run all the code in your `profile` file every time you run this script. E.g. @unnamedd will probably see his `neofetch` output (if this would be `fullOutput` mode command).

CleanShot 2020-11-15 at 17 04 17@2x

So I would like to understand why exactly this script doesn't work without login shell. It might be a $PATH issue which can be resolved explicitly. I tried locally without `-l` and it worked for me. What do you see in the output when you try dropping `-l` and use `fullOutput` mode?

browsing/website-screenshots.sh
```diff
@@ -0,0 +1,37 @@
...
+# @raycast.schemaVersion 1
+# @raycast.title Screenshot website from clipboard
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Screenshot website using sindresorhus/pageres-cli.
+# @raycast.packageName Internet
+# @raycast.needsConfirmation true
+# @raycast.currentDirectoryPath ~/Desktop
+
```
PitNikola, line 18: How about adding an icon? E.g. 🖼

browsing/website-screenshots.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -l
+
+# Dependecy: requires pageres (https://github.com/sindresorhus/pageres-cli)
+# Install via npm: `npm install --global pageres-cli`
+
+# Set viewport resolution of screenshots
+resolutions="1920x1080 390x844"
+
+# @raycast.schemaVersion 1
+# @raycast.title Screenshot website from clipboard
```
PitNikola, line 10:
```suggestion
# @raycast.title Screenshot Website from Clipboard
```

For consistency of the rest of Raycast commands I would suggest capitalising words in the title. We follow apple's human interface guidelines:

> Use title-case for push button titles. Capitalize every word except articles, coordinating conjunctions, and prepositions of four or fewer letters.

browsing/website-screenshots.sh
```diff
@@ -0,0 +1,37 @@
...
+# @raycast.mode silent
```
PitNikola, line 11: When I was running this command locally, I noticed that it takes quite a while to perform. What do you think about changing mode to `compact` here? `silent` usually works best for instant actions. Otherwise you don't get any feedback after an action.

PitNikola (changes requested):

browsing/website-screenshots.sh
crstauf: @PitNikola If I remove `-l` I get the error that the command doesn't exist.

I had a permissions issue which required me to create a different directory for global npm packages. With `-l` it should work for everyone without issue, including systems with a user-specific directory for packages.

browsing/website-screenshots.sh
crstauf: I chose `silent` for many reasons:
- it takes so long
- its not a blocking action
- the output directory is known
- there's a notification when completed outside of Raycast

crstauf: When choosing a mode I consider my behavior if the command were being run in Terminal, and what immediate use the script has. In this case, because of time required, I execute the command and go off to do something else, and come back to Terminal or where the files were saved.

commands/browsing/website-screenshots.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Set viewport resolution of screenshots
+resolutions="1920x1080 390x844"
+
+# @raycast.schemaVersion 1
+# @raycast.title Screenshot Website from Clipboard
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Screenshot website using sindresorhus/pageres-cli.
```
thomaspaulmann, line 14: Let's make this more user-friendly to read.
```suggestion
# @raycast.description Takes a screenshot of the Website from your Clipboard and saves it to the Desktop.
```

commands/browsing/website-screenshots.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Dependency: requires pageres (https://github.com/sindresorhus/pageres-cli)
+# Install via npm: `npm install --global pageres-cli`
+
+# Set viewport resolution of screenshots
+resolutions="1920x1080 390x844"
+
+# @raycast.schemaVersion 1
+# @raycast.title Screenshot Website from Clipboard
+# @raycast.mode silent
```
thomaspaulmann, line 11: I would recommend using the `compact` mode since it's a long-running network request. To be fair, our documentation lacks and we should make it clear when to use which mode.

commands/browsing/website-screenshots.sh
```diff
@@ -0,0 +1,38 @@
...
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Screenshot website using sindresorhus/pageres-cli.
+# @raycast.packageName Internet
+# @raycast.needsConfirmation true
```
thomaspaulmann, line 16: Why requiring confirmation?

commands/browsing/website-screenshots.sh
```diff
@@ -0,0 +1,38 @@
...
+
+if [ -z ${resolutions+x} ]; then
+	echo "Viewport resolutions are undefined.";
+	exit 0
+fi
+
+regex='(https?)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]'
+clipboard=$(pbpaste)
+
+if [[ $clipboard =~ $regex ]]; then
```
thomaspaulmann, line 33: Seems like a common pattern to detect URLs in scripts. Seems worth to keep that in mind for inputs. CC @PitNikola

commands/browsing/website-screenshots.sh
```diff
@@ -0,0 +1,38 @@
...
+	echo "Viewport resolutions are undefined.";
+	exit 0
+fi
+
+regex='(https?)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]'
+clipboard=$(pbpaste)
+
+if [[ $clipboard =~ $regex ]]; then 
+	output=$(pageres --crop "$clipboard" $resolutions)
+	echo "Saved screenshot(s) to Desktop"
```
thomaspaulmann, line 35: Can we pick appropriate singular or plural? What does it depend on?

thomaspaulmann (changes requested):
I left some suggestions to improve the script. Could you also rebase to avoid any merge commits on master? Thanks a lot.

commands/browsing/website-screenshots.sh
crstauf: I think this should be left up to the Script Command author. If a user wants to change it, they can. Decisions have been made that expect the user to make changes if they need to, and I think that should be the case here.

In addition, I chose `silent` because of how long it takes. I don't want to keep Raycast open while it runs; just start the command, and then I can find the files on my Desktop when I'm ready to use them. Easy.

commands/browsing/website-screenshots.sh
crstauf: To prompt the user to double-check their clipboard contents are correct.

commands/browsing/website-screenshots.sh
crstauf: How many URLs are in clipboard, how many resolutions are specified.

thomaspaulmann (changes requested):
As discussed in Slack, the `mode` and `needsConfirmation` is breaking our interface guidelines. Please adjust to make it easy for users to adopt this script command. Thanks!

thomaspaulmann (approved):
Thanks for adjusting the `mode` and `needsConfirmation` to fit our UX guidelines.

PR #69, opened 2020-11-13
Add Script Command: carbon-now

crstauf:
[Weirdness](https://github.com/raycast/script-commands/compare/master...crstauf:add/carbon-now?expand=1#diff-3f31d6bf693302081eff11356372852a65c8c3c7b7fb8235e691f965b2d5e709R13) as mentioned in [Slack](https://raycastcommunity.slack.com/archives/C01D1B36344/p1605247189239000) and https://github.com/raycast/script-commands/pull/68#issue-520381268:

> due to permissions troubles, i was forced to create a new global directory for npm packages… while a command works from my Terminal, i can’t get the same command to work from a Script Command… i’ve tried everything i know but not working. can someone with more knowledge on this please advise?


crstauf:

commands/developer-utils/create-image-of-clipboard.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Dependency: carbon-now-cli (https://github.com/mixn/carbon-now-cli)
+# Install via npm: `npm i -g carbon-now-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Create image of clipboard contents
```
thomaspaulmann, line 7: Please use title-cased strings for the title.
```suggestion
# @raycast.title Create Image of Clipboard Contents
```

commands/developer-utils/create-image-of-clipboard.sh
```diff
@@ -0,0 +1,23 @@
...
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Create beautiful image of clipboard contents (typically code).
+# @raycast.packageName Developer Utilities
+# @raycast.needsConfirmation true
```
thomaspaulmann, line 13: Why requiring confirmation for this script? Doesn't seem harmful.

commands/developer-utils/create-image-of-clipboard.sh
```diff
@@ -0,0 +1,23 @@
...
+# @raycast.currentDirectoryPath ~/Desktop
+
+if ! command -v carbon-now &> /dev/null; then
+	echo "carbon-now command is required (https://github.com/mixn/carbon-now-cli).";
+	exit 1;
+fi
+
+carbon-now --from-clipboard -h
+
+echo "Image saved to Desktop"
```
thomaspaulmann, line 23: I wonder if it makes sense to reveal the downloaded image directly. What do you think?

commands/developer-utils/create-image-of-clipboard.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Dependency: carbon-now-cli (https://github.com/mixn/carbon-now-cli)
+# Install via npm: `npm i -g carbon-now-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Create image of clipboard contents
+# @raycast.mode silent
```
thomaspaulmann, line 8: Let's use the `compact` mode in this case. Since it's a long-running network operation, closing Raycast and don't have a loading indicator doesn't seem good.
```suggestion
# @raycast.mode compact
```

commands/developer-utils/create-image-of-clipboard.sh
thomaspaulmann: You can specify the name of the downloaded file with the `--target` argument. Maybe use something along the lines of `carbon-{Date}.png`.

thomaspaulmann (changes requested):
Very nice addition! I used `carbon-now` to share the first script commands on Twitter :-) 

I added a few suggestions that I think would make the command more user-friendly. Let me know what you think. Also, could you please rebase your branch? We like to `rebase & merge` to keep a linear history. Thanks.

commands/developer-utils/create-image-of-clipboard.sh
crstauf: I mentioned elsewhere, whenever clipboard contents are used, I require confirmation, to cause the user to consider their clipboard's content.

Additionally, I ask for confirmation when files are being generated or altered.

commands/developer-utils/create-image-of-clipboard.sh
crstauf: I mentioned this elsewhere, that I try to replicate my behavior with terminal. Execute the command, go off and do something else, then grab the files when needed.

The general thinking seems also to be that if a user needs to alter the script in their local environment to suit their needs or system, it's okay to expect them to.

commands/developer-utils/create-image-of-clipboard.sh
crstauf: > reveal the downloaded image directly

I thought about that, and decided better to not interrupt the user, and let them use the image at their discretion. I don't always need it immediately, and am instead generating it for a blog post later.

> use something along the lines of `carbon-{Date}.png`

Smart; I'll add that.

commands/developer-utils/create-image-of-clipboard.sh
thomaspaulmann: The scripts in this repository should follow UX principles and all look and feel the same without altering them. In this case, it should be `compact` to properly show a loading indicator. Otherwise the user steps in the dark and doesn't know what happens with the command. You can still use Raycast while the script is running but it shows you a little toast on the bottom left corner to signal that there is something happening in the background.

This is actually more aligned with the `Terminal` where you also see that the command is still running.

thomaspaulmann: Please adopt the `mode` to adjust to the other scripts.

commands/developer-utils/create-image-of-clipboard.sh
thomaspaulmann: That doesn't align with our guideline on when to use the confirmation. This flag should only be used if there is a non-reversible action that get's executed, e.g. some file is getting deleted. 

In the case of this command, nothing harmful happens and the title describes that it uses the clipboard. Can you please adopt? Thanks.

commands/developer-utils/create-image-of-clipboard.sh
thomaspaulmann: Yeah, I wasn't sure about opening the image. Sounds like a good idea to be passive in this case. Thanks for adding the date!

thomaspaulmann (changes requested):
Please follow our guidelines on how to use `mode` and `needsConfirmation`. It's important that these properties are used consistently to make the commands feel aligned with each other.

crstauf:
I believe I have followed the guidelines...?

I chose `silent` because the user experience is best for the command to "run in background", and `needsConfirmation` because I want to show "confirmation alert dialog before running the script" to ensure proper clipboard content.

fe9lix:
I like the idea of that command, however I see a couple of issues in its current form:

- this dependency is really heavy: it downloads entire chromium (!) as part of its NPM dependencies...
- it doesn't work for me, not even when i directly use the plain CLI version of carbon-now (some browser connection errors, etc.); while this wouldn't be an argument in itself to reject it, chances are that other people may run into the same issue since i have an otherwise working node/npm setup on my machine

This would be cool if it there was an API where you send some (non-sensitive) code snippet to and get some binary back with the image but this setup is much more involved here since it seems to run it through browser snapshotting.

@crstauf This one might be better suited for your own personal usage but not so much for the public repo in its current form since it may not work out of the box, not even with the dependency installed.

crstauf:
It is unfortunate that the CLI command downloads Chromium... the `-h` (headless) option skips its use, but nevertheless is part of the CLI command.

PR #70, opened 2020-11-13
Add Script Command: download-url

crstauf:
[Weirdness](https://github.com/raycast/script-commands/compare/master...crstauf:add/download-url?expand=1#diff-479aaa3e7b2f8b9056dc772d91cd5b1aeff0e57519fa4aedb625f1699c8cd366R13) as mentioned in [Slack](https://raycastcommunity.slack.com/archives/C01D1B36344/p1605247189239000) and https://github.com/raycast/script-commands/pull/68#issue-520381268:

> due to permissions troubles, i was forced to create a new global directory for npm packages… while a command works from my Terminal, i can’t get the same command to work from a Script Command… i’ve tried everything i know but not working. can someone with more knowledge on this please advise?


browsing/download-url.sh
```diff
@@ -0,0 +1,29 @@
...
+# @raycast.title Download URL in clipboard
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Download URL in clipboard.
+# @raycast.packageName Developer Utilities
+# @raycast.needsConfirmation true
+# @raycast.currentDirectoryPath ~/Desktop
+
+cmnd=$(npm config get prefix)/bin/download
```
unnamedd, line 13: Instead of using that `download`, why not use `cURL` do to the job? The chances to use the computer has it, is bigger than this one and reduce one step to make this script command work.

browsing/download-url.sh
```diff
@@ -0,0 +1,29 @@
...
+if ! command -v $cmnd &> /dev/null; then
+	echo "download command is required (https://github.com/kevva/download-cli).";
+	exit 1;
+fi
+
+clipboard=$(pbpaste)
+regex='(https?)://[-A-Za-z0-9\+&@#/%?=~_|!:,.;]*[-A-Za-z0-9\+&@#/%=~_|]'
+
+if [[ $clipboard =~ $regex ]]; then 
+	"$cmnd" -o ~/Desktop $clipboard
```
unnamedd, line 24: What will be the name of the file downloaded here? The full URL or just the of it? (In this case the filename + extension)

crstauf: Would be the filename and extension.

browsing/download-url.sh
crstauf: Because with `curl`, filename has to be specified; `download` handles that on its own. Perhaps I'll submit a separate Script Command for quickly running `curl`.

unnamedd: So then I guess we can do this in this PR instead of a second one. I don't see reasons to have two different script commands one to use `download-cli` and one to use `cURL`. To use cURL, we will have to work a bit more, but at least it will solve the problem and no one will have to install any extra thing. What do you think?

browsing/download-url.sh
unnamedd: I'm sure we can manage that by ourselves.

browsing/download-url.sh
crstauf: If someone submits a PR for `curl` that has the same behavior as `download`, I'll gladly close this PR in favor of that one.

crstauf:

commands/browsing/download-url.sh
```diff
@@ -0,0 +1,30 @@
...
+# Install via npm: `npm install --global download-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Download URL in clipboard
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Download URL in clipboard.
+# @raycast.packageName Developer Utilities
+# @raycast.needsConfirmation true
```
thomaspaulmann, line 13: Let's remove this. I don't think we need a confirmation for this script.

commands/browsing/download-url.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Dependency: emoj (https://github.com/kevva/download-cli)
+# Install via npm: `npm install --global download-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Download URL in clipboard
```
thomaspaulmann, line 7: Title case :-)
```suggestion
# @raycast.title Download URL in Clipboard
```

commands/browsing/download-url.sh
```diff
@@ -0,0 +1,30 @@
...
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Download URL in clipboard.
```
thomaspaulmann, line 11: What does it download? The sources? The entire page? Just HTML? Could you please clarify?

commands/browsing/download-url.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Dependency: emoj (https://github.com/kevva/download-cli)
+# Install via npm: `npm install --global download-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Download URL in clipboard
+# @raycast.mode silent
```
thomaspaulmann, line 8: This should be a `compact` script since it does a long-running network request.

```suggestion
# @raycast.mode compact
```

thomaspaulmann (changes requested):
Could you clarify what this command downloads? I also added some suggestions for title, mode and confirmation.

Thanks for another contribution, yeah!

commands/browsing/download-url.sh
crstauf: I request confirmation whenever the clipboard is in use, as a prompt to the user to check the clipboard's contents. Also whenever files are being generated.

commands/browsing/download-url.sh
crstauf: I specifically choose `silent` in these cases because I'd rather set it running and leave it, and come back to it later.

commands/browsing/download-url.sh
crstauf: Downloads whatever is present at the URL:

|URL|File type|
|---|---|
|https://github.com|HTML|
|https://github.com/styles.css|CSS|
|https://github.com/scripts.js|JavaScript|
|https://github.com/logo.png|PNG|

commands/browsing/download-url.sh
thomaspaulmann: As discussed in Slack, please make sure to use `compact` to follow other the UX of other commands in this repository. Thanks

thomaspaulmann (changes requested):
Please adjust the `mode` to be `compact` and remove `needsConfirmation` to align the command to the others in this repository. Thanks.

fe9lix:
@crstauf What does this script do exactly or when would you use it? If it downloads the passed URL to a certain location, let's use curl instead of the download-url package here.

crstauf:
> let's use `curl` instead of the `download-url` package

I'm unable to. Closing.

PR #80, opened 2020-11-16
Add Script Command: strip-indent

crstauf:
## Description

Strip leading whitespace from each line in a string. The line with the least number of leading whitespace, ignoring empty lines, determines the number to remove.

## Type of change

- [x] New script command

## Screenshot

code-with-indent
Screen Shot 2020-11-16 at 3 46 01 PM
code-without-indent

https://cloudup.com/clcPy2PVtVG

Requires [sindresorhus/strip-indent-cli](https://github.com/sindresorhus/strip-indent-cli/).

## Checklist


commands/developer-utils/strip-indent.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Dependency: requires strip-indent (https://github.com/sindresorhus/strip-indent-cli)
+# Install via npm: `npm install --global strip-indent-cli`
+
+# @raycast.schemaVersion 1
+# @raycast.title Shift clipboard contents left
```
thomaspaulmann, line 7: Title case please :-) Also, I didn't get the title. How about that?

```suggestion
# @raycast.title Strip Indent from Clipboard Contents
```

commands/developer-utils/strip-indent.sh
```diff
@@ -0,0 +1,26 @@
...
+# @raycast.mode silent
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Strip leading whitespace from each line in a string.
+# @raycast.packageName Developer Utilities
+# @raycast.needsConfirmation true
```
thomaspaulmann, line 13: No need for confirmation. Lets remove it.

thomaspaulmann (changes requested):
Mmh... This are two regexes with some math (https://github.com/jamiebuilds/min-indent/blob/master/index.js, https://github.com/sindresorhus/strip-indent/blob/master/index.js). Do we really need to pull in some CLI for this? No offence but I think you could simply write that with a plain script and it would be simpler to adopt from others.

commands/developer-utils/strip-indent.sh
crstauf: I prefer to have the user confirm their clipboard value is correct before proceeding.

commands/developer-utils/strip-indent.sh
crstauf: I'm concerned "Strip Indent from Clipboard Contents" may suggest all indents are removed, which isn't correct. Text and code editors commonly use the "shift" language to describe moving blocks of text left or right.

crstauf:
> Do we really need to pull in some CLI for this?

I do; if someone else can build something without a package, cool.

crstauf:
Closing this so someone else can recreate without a dependency.

PR #82, opened 2020-11-17
Add Script Command Template: set-screens-preset

crstauf:
## Description

Introduce a template that uses `jakehilborn/displayplacer` to create presets for system displays. I personally use it to change the orientation of my iMac between landscape and portrait, but much more complex presets can be created.

## Type of change

- [x] New script command

## Screenshot

Screen Shot 2020-11-17 at 9 29 53 AM

Requires `jakehilborn/displayplacer` dependency, as well as some config (thus submitting as template).

## Checklist

### DIFF
--- a/commands/system/set-screens-preset.template.sh
+++ b/commands/system/set-screens-preset.template.sh
@@ -20,6 +20,7 @@ settings=""
 # @raycast.author Caleb Stauffer
 # @raycast.authorURL https://github.com/crstauf
 # @raycast.description Apply preset screen settings.
+# @raycast.packageName System
 # @raycast.icon 🖥️
 
 if ! command -v displayplacer &> /dev/null; then
--- a/commands/system/set-screens-preset.template.sh
+++ b/commands/system/set-screens-preset.template.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Dependency: requires displayplacer (https://github.com/jakehilborn/displayplacer)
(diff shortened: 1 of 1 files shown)


commands/system/set-screens-preset.template.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Dependency: requires displayplacer (https://github.com/jakehilborn/displayplacer)
+# Install via Homebrew: `brew tap jakehilborn/jakehilborn && brew install displayplacer`
+
+##################################################################
+## Get display ID and settings by running `displayplacer list`, ##
+## find the command, and copy command to `settings` variable.   ##
+##                                                              ##
+## See https://github.com/jakehilborn/displayplacer for help.   ##
+##################################################################
```
thomaspaulmann, line 11: This seems like a lot to ask for to set up this command. We tend to have simple to set up and ready to use scripts in this repository.

thomaspaulmann (changes requested):
Thanks for adding another command but I don't think this is command is a great fit for this repository. We prefer ready-to-use commands that users can easily adopt and don't need to configure extensively outside of Raycast. That said, the command seems certainly useful for `displayplacer` users. Would there be any chance to minimise the setup costs?

crstauf:
> Would there be any chance to minimise the setup costs?

Not that I'm aware of.

Perhaps a user wants the functionality, but doesn't know it exists or how to do it... enter Raycast and this script template.

PR #85, opened 2020-11-19
Workflow improvements

unnamedd:
## Description

Fixes a small bug when the documents are generated (or regenerated) even without any changes on the script commands, we were committing the `extensions.json` with only the `lastUpdate` changed. 

The intention isn't make any commit using the Raycast Bot if we didn't explicitly made changes in two files: `commands/extensions.json` and `commands/README.md`.  This pull request is fixing that.

## Type of change

- [x] Toolkit change
            - Renamed `lastUpdate` key to `updatedAt`
- [x] Other:
            - Workflow update

## Screenshot

No apply

No apply

## Checklist

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -21,16 +21,7 @@ jobs:
 
     - name: Generate Script Commands Documentation
       run: |
-        make gen-docs
-        if `git status | grep -q "nothing to commit"`; then
-          exit 0;
-        else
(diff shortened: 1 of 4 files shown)


thomaspaulmann:
Thanks for making sure that the workflow runs cleanly. As discussed in Slack, it wasn't clear what the `updatedAt` property means. Might be worth to add some comments to make sure that it isn't misinterpreted.

PR #88, opened 2020-11-19
Add Search for Script Command

unnamedd:
## Description

A small script command to make possible the search of script commands in the GitHub repository.

## Type of change

- [x] New script command

## Screenshot

image

Dependency: Swift

## Checklist


japanese-goblinn:
Seems good. Don't you think to add a line above the URL with only a script name for better UX?

thomaspaulmann:
It isn't that "small" of a script anymore ;-) Cool addition to explore script commands in this repository. We might add something like this to Raycast but for the time-being it's a great addition to quickly find a script. We should definitely add parsing for URLs that you can open them via the Action Panel.

One minor change, I'm not sure if `Web Searches` is the right directory for this. Yes it's a search, but web searches tend to open the browser and are rather shortcuts to get from everywhere to a new tab with your query. Should we move it to `developer tools` or any other idea?

unnamedd:
@thomaspaulmann I had the same question when I added it in the "Web Searches", because it doesn't fit there, I completely agree with you, but also, it is not part of the developer utils, but at least, in the current moment, it fits there, but not completely... 

And yes, this script command has a short life and I will be very happy to see a better version of it built-in in Raycast!

ps: come on, it is smaller than the AirPods Battery! 😜

thomaspaulmann:
> ps: come on, it is smaller than the AirPods Battery! 😜

Haha yeah that is true. Most of the stuff are models. Would be cool if handling JSONs could be simpler in Swift. I guess one of the downsides of a typesafe language 🤷

PR #94, opened 2020-11-20
Set Spotify URL script executable

crstauf:
## Description

Set `spotify-now-playing-url` script as executable.

## Type of change

## Checklist


thomaspaulmann:
Thanks for fixing! We should add some CI to check if new scripts are executable.

unnamedd:
> Thanks for fixing! We should add some CI to check if new scripts are executable. 

Soon! ;)

PR #99, opened 2020-11-22
Add script command: Define a word

jesse-c:
## Description

It adds a script command for looking up the definition of a word using the macOS dictionary.

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

Screen Shot 2020-11-22 at 1 43 29 pm

N/A
## Checklist

### DIFF
--- a/commands/system/define-word.swift
+++ b/commands/system/define-word.swift
@@ -15,13 +15,13 @@
 
 import Foundation
 
-if CommandLine.argc != 2 {
-  print("Must pass 1 word to define")
-} else {
+if CommandLine.argc > 1 {
   let argument = CommandLine.arguments[1].lowercased()
 
   // https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/DictionaryServicesProgGuide/access/access.html#//apple_ref/doc/uid/TP40006152-CH5-SW3
   let result = DCSCopyTextDefinition(nil, argument as CFString, CFRangeMake(0, argument.count))?.takeRetainedValue() as String?
 
   print(result ?? "No definition found")
+} else {
+  print("Must pass 1 word to define") 
 }
--- a/commands/system/define-word.swift
+++ b/commands/system/define-word.swift
@@ -1,0 +1,27 @@
+#!/usr/bin/swift
+
+// Required parameters:
+// @raycast.schemaVersion 1
+// @raycast.title Define Word
+// @raycast.author Jesse Claven
+// @raycast.authorURL https://github.com/jesse-c
+// @raycast.description Define a word using the built-in dictionary/dicionaries.
+// @raycast.packageName System
+// @raycast.mode fullOutput
+//
+// Optional parameters:
+// @raycast.icon 🗣
+// @raycast.argument1 { "type": "text", "placeholder": "Word (e.g. isthmus)" }
+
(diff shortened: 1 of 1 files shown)


commands/system/define-word.swift
```diff
@@ -0,0 +1,27 @@
...
+if CommandLine.argc != 2 {
+  print("Must pass 1 word to define")
+} else {
+  let argument = CommandLine.arguments[1].lowercased()
+
+  // https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/DictionaryServicesProgGuide/access/access.html#//apple_ref/doc/uid/TP40006152-CH5-SW3
+  let result = DCSCopyTextDefinition(nil, argument as CFString, CFRangeMake(0, argument.count))?.takeRetainedValue() as String?
+
+  print(result ?? "No definition found")
+}
```
unnamedd, line 27: I suggest you to check if the amount of arguments if you received is greater than 1, then you will ensure that you can ask for the argument in the second position (`[1]`) without receive an `index out of bound` error:

```suggestion
if CommandLine.argc > 1 {
  let argument = CommandLine.arguments[1].lowercased()

  // https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/DictionaryServicesProgGuide/access/access.html#//apple_ref/doc/uid/TP40006152-CH5-SW3
  let result = DCSCopyTextDefinition(nil, argument as CFString, CFRangeMake(0, argument.count))?.takeRetainedValue() as String?

  print(result ?? "No definition found")
} else {
  print("Must pass 1 word to define") 
}
```

unnamedd (changes requested):

commands/system/define-word.swift
jesse-c: That's a fair suggestion! Thanks.

commands/system/define-word.swift
```diff
@@ -0,0 +1,27 @@
...
+
+if CommandLine.argc > 1 {
+  let argument = CommandLine.arguments[1].lowercased()
+
+  // https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/DictionaryServicesProgGuide/access/access.html#//apple_ref/doc/uid/TP40006152-CH5-SW3
+  let result = DCSCopyTextDefinition(nil, argument as CFString, CFRangeMake(0, argument.count))?.takeRetainedValue() as String?
+
+  print(result ?? "No definition found")
+} else {
+  print("Must pass 1 word to define")
```
thomaspaulmann, line 26: This should never happen 🤞

thomaspaulmann (approved):
Cool one! Just an heads up, we do plan to add support for definitions sooner or later. For now we can keep the script command but this might be replaced by some built-in solution.

commands/system/define-word.swift
unnamedd: It might happen if you execute the script manually via terminal, not via Raycast, but it's never bad to be safer, right?

thomaspaulmann: Yes, always better to be safe!

jesse-c:
Happy to see the script removed once there's some built-in functionality!

PR #101, opened 2020-11-22
Add Script Commands: largest-cpu-process, largest-ram-process

crstauf:
## Description

Please write a short summary for this change. If it's a new script command, explain what it does.

## Type of change

- [x] New script command

## Screenshot

Screen Shot 2020-11-22 at 12 35 46 PM

- `ps` command
- `awk` command

## Checklist

### DIFF
--- a/commands/system/largest-cpu-process.sh
+++ b/commands/system/largest-cpu-process.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Largest CPU Process
+# @raycast.mode inline
+# @raycast.refreshTime 1m
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Report process with largest system CPU usage.
+# @raycast.packageName System
+# @raycast.icon 📈
+
+output=$(ps -exco pcpu,comm | sort -k 1 -n -r | head -n 1)
+percentage=$(echo -e "$output" | awk -F " " '{ st = index($0," ");print $1}')
+
+# Using awk like in largest-ram-consumer doesn't work, so do this instead.
+cmd=${output/$percentage/}
+
+# Remove leading whitespace...?
+cmd=$(echo "$cmd" | awk '{$1=$1};1')
+
+echo "${cmd} - ${percentage}%"
--- a/commands/system/largest-ram-process.sh
+++ b/commands/system/largest-ram-process.sh
@@ -7,7 +7,7 @@
 # @raycast.refreshTime 1m
 # @raycast.author Caleb Stauffer
 # @raycast.authorURL https://github.com/crstauf
-# @raycast.description Report process with greatest system RAM usage.
+# @raycast.description Report process with largest system RAM usage.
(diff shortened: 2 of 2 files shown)


commands/system/largest-ram-process.sh
```diff
@@ -0,0 +1,20 @@
...
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Report process with greatest system RAM usage.
+# @raycast.packageName System
+# @raycast.icon 📈
+
+output=$(ps -exco rss,comm | sort -k 1 -n -r | head -n 1)
+bytes=$(echo "$output" | awk -F " " '{ st = index($0," ");print $1}')
+cmd=$(echo "$output" | awk -F " " '{ st = index($0," ");print substr($0,st+1)}')
+
+gigabytes=$(echo "scale=3; $bytes / 1024 / 1024" | bc -l)
```
thomaspaulmann, line 18: I assume `scale` are the digits after the decimal point? Let's make it 1. This should be enough and aligns with how the Activity Monitor displays it.
```suggestion
gigabytes=$(echo "scale=1; $bytes / 1024 / 1024" | bc -l)
```

commands/system/largest-cpu-process.sh
```diff
@@ -0,0 +1,23 @@
...
+# @raycast.icon 📈
+
+output=$(ps -exco pcpu,comm | sort -k 1 -n -r | head -n 1)
+percentage=$(echo -e "$output" | awk -F " " '{ st = index($0," ");print $1}')
+
+# Using awk like in largest-ram-consumer doesn't work, so do this instead.
+cmd=${output/$percentage/}
+
+# Remove leading whitespace...?
+cmd=$(echo "$cmd" | awk '{$1=$1};1')
```
thomaspaulmann, line 21: So does this remove the leading whitespace or not? 🙃

commands/system/largest-ram-process.sh
thomaspaulmann: I would also suggest to show process that consume less than 1 GB of RAM as megabytes, e.g. 834.6 MB. This would align with the Activity Monitor. 

Something along these lines:
```suggestion
if [ $bytes -le 1024 ]; then
  scaled_bytes=...
el
  scaled_bytes=...
fi

formatted_bytes=$(echo "scale=1; $scaled_bytes" | bc -l)
```
  formatted_bytes=$(echo "scale=1; $bytes / 1024 " | bc -l)
  formatted_bytes=$(echo "scale=1; $bytes / 1024 / 1024" | bc -l)

```

thomaspaulmann (changes requested):
Nice one! I added some suggestions to improve the visual representation of the numbers.

commands/system/largest-cpu-process.sh
crstauf: It does, I don't understand why it's needed here and not in the RAM one.

commands/system/largest-ram-process.sh
crstauf: @thomaspaulmann Updated; set to one decimal place for megabytes and two decimal places for gigabytes (to match Activity Monitor https://cloudup.com/coGkuBW6PUb).

PR #107, opened 2020-11-23
Add Script Command: pa11y

crstauf:
## Description

Add Script Command to run `pa11y` web accessibility tool, and generate an HTML report of the errors on the user's Desktop.

## Type of change

- [x] New script command

## Screenshot

Kapture 2020-11-23 at 15 08 01

## Checklist


crstauf:
Expect this PR to be rejected due to use of dependencies, but submitting nevertheless.

fe9lix:
Yeah, you guessed it already: It's something rather specialized + external dependency requirement, which doesn't make it an ideal candidate for sharing. (We can revisit such scripts, once we introduced a better way in Raycast for handling script requirements.)

PR #118, opened 2020-11-27
Fix RecordSimulator script, and use arguments

maximkrouk:
## Description

I'm not sure that it was working perfectly, @japanese-goblinn noticed, that there were some issues (Maybe it was because of the API, where output **filename** was taken from the clipboard). Now the script uses arguments. Also it will show a message if the file already exists and will reveal it in the finder.

## Screenshot

<p>

Screen Shot 2020-11-27 at 11 44 17

Screen Shot 2020-11-27 at 11 44 38

Screen Shot 2020-11-27 at 11 44 55

Screen Shot 2020-11-27 at 11 45 22

Screen Shot 2020-11-27 at 11 45 36

Screen Shot 2020-11-27 at 11 44 55

Screen Shot 2020-11-27 at 11 46 05

Screen Shot 2020-11-27 at 11 46 18

</p>

## Checklist


shpakovski (approved):
👏

commands/developer-utils/record-simulator.swift
```diff
@@ -0,0 +1,35 @@
...
+// Optional parameters:
+// @raycast.author Maxim Krouk
+// @raycast.authorURL https://github.com/maximkrouk
+// @raycast.description Records simulator to Downloads folder
+// @raycast.needsConfirmation true
+// @raycast.icon 📱
+// @raycast.argument1 { "type": "text", "placeholder": "Filename" }
+// @raycast.currentDirectoryPath ~/Downloads
+
+import Foundation
```
unnamedd, line 18: Basically, you are creating a Swift script to execute a command in terminal like Raycast is already doing then, I need to ask: doesn’t make more sense to implement it completely in Bash instead of Swift? 

The huge benefit to develop in our programming language is also a weak point. Why? Because the Swift file will have to be compiled before executed and Raycast doesn’t have cache yet, because of that, the execution will take longer than needed.

commands/developer-utils/record-simulator.swift
```diff
@@ -0,0 +1,35 @@
...
+    let task = Process()
+    task.launchPath = "/bin/zsh"
+    task.arguments = ["-c", args.joined(separator: " ")]
+    task.launch()
+    task.waitUntilExit()
+    return task.terminationStatus
+}
+
+let fullName = CommandLine.arguments[1]
+    .replacingOccurrences(of: " ", with: #"\ "#)
```
unnamedd, line 31: This replacement can be done easily with shell, just doing that: `${1// /%20}`.

commands/developer-utils/record-simulator.swift
```diff
@@ -0,0 +1,35 @@
...
+    .appending(".mp4")
+
+shell("xcrun", "simctl", "io", "booted", "recordVideo", fullName)
+shell("open", "-R", fullName)
```
unnamedd, line 35: The execution of these commands here, can be done with no problems in shell and them will be executed really much faster than written in Swift.

unnamedd (changes requested):
Hey @maximkrouk,

I liked a lot your script in Swift, but there are some comments about a conversion of it to shell. Please, have a look and if you decide to dive into those cold waters, I can you there.

shpakovski:
@maximkrouk Sorry for ignorance but when I quit the Simulator, the script is still running:

image

When I cancel running explicitly, the recorded `.mp4` cannot be open by the QuickTime.

Could you please suggest how to use it?

maximkrouk:
@unnamedd 
Maybe I will do it later but for now, it's probably better to have a convenient argument-based Swift script, than a bash script with a strange clipboard API. Also Swift kinda handles errors for me, so the output and script readability is better out-of-the-box. I totally agree that it's a bit slower, it should compile, import Foundation, call zsh to execute commands, that is all true, but anyway it works fine, so I'll dive into error handling in bash a bit later 🙂

commands/developer-utils/record-simulator.swift
shpakovski: I must agree with @unnamedd Swift doesn’t provide anything special here but adds compilation overhead 😅

maximkrouk:
@shpakovski, You should probably cancel the recording while the simulator is open (interrupt the script from the Raycast UI)

maximkrouk:
Oh, got you, something goes wrong, I'll take a look in the evening, don't merge yet 😢

unnamedd:
> @unnamedd
> Maybe I will do it later but for now, it's probably better to have a convenient argument-based Swift script, than a bash script with a strange clipboard API. Also Swift kinda handles errors for me, so the output and script readability is better out-of-the-box. I totally agree that it's a bit slower, it should compile, import Foundation, call zsh to execute commands, that is all true, but anyway it works fine, so I'll dive into error handling in bash a bit later 🙂

I will have to disagree with you in one specific point: you are not doing any error handling in Swift. The output, will be the same in shell or in Swift, but you got an overhead here: you have the compilation. And it is not “bit slower”, compared to shell it is veeery much slower. 

Does that means that no Swift script command should be submitted? Definitely not. Means that we need to take in count what we want to achieve to keep the simplicity and the speed. In the current moment, there aren’t many benefits to write this script command in Swift, if it can be rewritten in shell with almost no much cost and I can help you.

maximkrouk:
Well, your truth, it was easy

maximkrouk:
Btw it's working now, feel free to merge <3

shpakovski:
I can’t make it work, the file is unplayable:

image

Maybe this is my setup… Could someone else check?

maximkrouk:
May you run thes commands in terminal? If it works try to check if you pulled the latest (bash) version
```bash
cd ~/Downloads
xcrun simctl io booted recordVideo Output.mp4
```

fe9lix (commented):
Hey @maximkrouk! In addition to what was mentioned above re the Swift version (it also didn't work on my machine), I was wondering why calling the commands through /bin/zsh first and not directly through Process()? 

Regarding the sh version:
- Have you tested if it works with spaces in the argument/filename?
- You can take advantage of the currentDirectoryPath parameter (so you're basically already in the downloads folder...)

maximkrouk:
> @fe9lix I was wondering why calling the commands through /bin/zsh first and not directly through Process()?
I needed to call 2 commands so I decided to generalize it a bit, I think it's just similar in this case, a little code duplication or unnecessary generalization

- Spaces should be escaped with slashes yet
- I tried, but got the same issue - `.mp4` output is unplayable

fe9lix (approved):

PR #127, opened 2020-11-30
Add Script Command Template: add-ssh-keys

crstauf:
## Description

AppleScript to open a Terminal window and run commands to add multiple keys to the SSH agent.

## Type of change

- [x] New script command

## Screenshot

Screen Shot 2020-11-29 at 11 40 30 PM

## Checklist

### DIFF
--- a/commands/developer-utils/add-ssh-keys.template.applescript
+++ b/commands/developer-utils/add-ssh-keys.template.applescript
@@ -1,0 +1,32 @@
+#!/usr/bin/osascript
+
+#######################################################
+## Set keyPaths array with paths to SSH keys to add. ##
+#######################################################
+
+# Array of paths to SSH keys.
+# Example: set keyPaths to { "~/.ssh/id_ed25519" }
+set keyPaths to {}
+
+# @raycast.title Add SSH keys
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Add SSH keys.
+
+# @raycast.icon 🔑
+# @raycast.mode silent
+# @raycast.schemaVersion 1
+
+on implode( array, delimiter )
+	set AppleScript's text item delimiters to delimiter
+	set implodedArray to array as string
+	set AppleScript's text item delimiters to ""
+	return implodedArray
+end implode
+
+set cmd to "clear; ssh-add " & implode( keyPaths,"; clear; ssh-add " )
+
+tell application "Terminal"
+	do script cmd
+	activate
+end tell
--- a/commands/developer-utils/add-ssh-keys.template.applescript
+++ b/commands/developer-utils/add-ssh-keys.template.applescript
@@ -8,7 +8,7 @@
 # Example: set keyPaths to { "~/.ssh/id_ed25519" }
 set keyPaths to {}
 
-# @raycast.title Add SSH keys
+# @raycast.title Add SSH Keys
 # @raycast.author Caleb Stauffer
 # @raycast.authorURL https://github.com/crstauf
 # @raycast.description Add SSH keys.
--- a/commands/developer-utils/add-ssh-keys.template.applescript
+++ b/commands/developer-utils/add-ssh-keys.template.applescript
@@ -15,6 +15,7 @@ set keyPaths to {}
 
 # @raycast.icon 🔑
 # @raycast.mode silent
+# @raycast.packageName Developer Utilities
 # @raycast.schemaVersion 1
 
 on implode( array, delimiter )
--- a/commands/developer-utils/add-ssh-keys.template.applescript
+++ b/commands/developer-utils/add-ssh-keys.template.applescript
@@ -8,10 +8,10 @@
 # Example: set keyPaths to { "~/.ssh/id_ed25519" }
 set keyPaths to {}
 
-# @raycast.title Add SSH Keys
+# @raycast.title Add SSH Keys (with Terminal)
 # @raycast.author Caleb Stauffer
 # @raycast.authorURL https://github.com/crstauf
-# @raycast.description Add SSH keys.
+# @raycast.description Add one or multiple SSH keys to your SSH agent (with Terminal).
 
 # @raycast.icon 🔑
 # @raycast.mode silent


commands/developer-utils/add-ssh-keys.template.applescript
```diff
@@ -0,0 +1,32 @@
...
+
+#######################################################
+## Set keyPaths array with paths to SSH keys to add. ##
+#######################################################
+
+# Array of paths to SSH keys.
+# Example: set keyPaths to { "~/.ssh/id_ed25519" }
+set keyPaths to {}
+
+# @raycast.title Add SSH keys
```
PitNikola, line 11:
```suggestion
# @raycast.title Add SSH Keys
```

PitNikola (changes requested):
Could you please explain in more details the use case for this script?

crstauf:
> Could you please explain in more details the use case for this script?

@PitNikola Sorry, I'm not sure how to add further details to "add multiple keys to the SSH agent".

Do you have a specific question I can answer regarding the scripts functionality?

PitNikola:
While it says "adds SSH keys", what it actually does is "opens terminal and adds SSH keys". What do you think about renaming title to be more specific about Terminal part?

As to my question about the use case, I would expect the answer along these lines: "I often run ssh agent and need to add bunch of ssh keys every time so with this command I just start terminal session with proper keys straight from Raycast". This way it would be much easier to understand why it's useful, especially for those who's not dealing with ssh-agent on regular basis.

PitNikola (changes requested):
Also let's add `packageName`.

crstauf:
> While it says "adds SSH keys", what it actually does is "opens terminal and adds SSH keys". What do you think about renaming title to be more specific about Terminal part?

@PitNikola By this logic, then all the scripts that open directories or websites should also be renamed: "Open website.com in Browser" and "Open Applications in Finder".

> I would expect the answer along these lines: "I often run ssh agent and need to add bunch of ssh keys every time so with this command I just start terminal session with proper keys straight from Raycast".

Thank you for the assist; I'm also confused: if you knew what it was for, why did you ask for more detail?

PitNikola:
> By this logic, then all the scripts that open directories or websites should also be renamed: "Open website.com in Browser" and "Open Applications in Finder"

Not really the same. For browser you have single default browser and we just run command `open` letting system to decide. Same for Finder. In your script you specifically ask to open "Terminal". Plus it's much more specific niche use case than opening URLs and folders.

> Thank you for the assist; I'm also confused: if you knew what it was for, why did you ask for more detail?

Because I did some search. By providing more context and not assuming that reviewers will know in details such such use cases, you will help us making reviews faster.

crstauf:
> you specifically ask to open "Terminal"

@PitNikola That makes sense, thank you for explaining. 

Is it possible to change the default terminal in macOS? If so, is there a way in AppleScript to have the system open and use that terminal?

PitNikola:
I don't think there is such a thing as default Terminal, but I'm not 100% sure. Feel free to do some researching.

crstauf:
> I don't think there is such a thing as default Terminal

@PitNikola Wouldn't the application "Terminal" be macOS default terminal?

> Feel free to do some researching.

I have done, and have not found any instructions to set macOS default terminal to something other than Terminal. Because that's the case, I don't see a need to add "using Terminal" to the script command's title.

PitNikola (changes requested):
> Wouldn't the application "Terminal" be macOS default terminal?

I mean there is no such notion as default terminal in the same way default browser exist. So when you write `tell application "Terminal"`, you're actually referring Apple's Terminal application.

>  I don't see a need to add "using Terminal" to the script command's title.

Huge amount of devs on macOS use iTerm. Making the title explicit will help with avoiding confusion. Also I would make description a bit more comprehensive, right now it just repeats the title.

crstauf:
Thanks for the discussion, @PitNikola; adjusted.

commands/developer-utils/add-ssh-keys.template.applescript
```diff
@@ -0,0 +1,33 @@
...
+
+#######################################################
+## Set keyPaths array with paths to SSH keys to add. ##
+#######################################################
+
+# Array of paths to SSH keys.
+# Example: set keyPaths to { "~/.ssh/id_ed25519" }
+set keyPaths to {}
+
+# @raycast.title Add SSH Keys (with Terminal)
```
PitNikola, line 11:
```suggestion
# @raycast.title Add SSH Keys to Terminal
```

crstauf: This is not an accurate title... the keys are not added to Terminal, they're added to the machine's SSH agent.

What's wrong with "(with Terminal)"?

PitNikola: Why it's not accurate? You're adding ssh keys to current Terminal session. Another alternative is `in Terminal`. I would avoid brackets unless it's really necessary (this one doesn't seem like the case) because it makes command less human readable.

crstauf: > Why's it not accurate.

See previous statement (emphasis added):

> the keys are not added _to_ Terminal

---

> You're adding ssh keys to current Terminal session

If (I understand correctly that) a Terminal session exists only as long as the Terminal window is open, this is not accurate. Once I add the keys to the `ssh-agent`, I can close Terminal and re-open, and my SSH keys are still on the agent.

> Another alternative is `in Terminal`

This is not accurate either, as the keys added to `ssh-agent` are available outside of Terminal. The most accurate is the form I have provided: `with Terminal`.

> I would avoid brackets

Cool; you don't have to use brackets in your script titles. I'm choosing to use brackets as its unnecessary information, and you've requested language to indicate use of Terminal (even thought it'd be clear to the user after the first use).

> makes command less human readable

I agree, but because you've requested language about Terminal and I'm trying to be kind, this is the form I am proposing as it seems most appropriate with the information it provides. See below explanation of parentheses:

> Two parentheses, ( ), are generally paired and used to mark off explanatory or qualifying remarks in writing.

That's exactly what "with Terminal" is doing... explaining.

PitNikola: > If (I understand correctly that) a Terminal session exists only as long as the Terminal window is open, this is not accurate. Once I add the keys to the ssh-agent, I can close Terminal and re-open, and my SSH keys are still on the agent.

Got it, makes sense.

> Cool; you don't have to use brackets in your script titles.

I'm afraid this kind of argument will not work well here. While we welcome contributions in this repository, we want to keep this place tidy and consistent because new contributors will come and follow existing examples. This particular case is rather borderline and is not worth fighting, but seeing so much resistance in PR reviews makes it much harder to collaborate.

> This is not accurate either, as the keys added to ssh-agent are available outside of Terminal. The most accurate is the form I have provided: with Terminal.

I would argue here because the sentence "adding ssh keys in Terminal application" is absolutely valid. But as I mentioned before, it's not too big a deal.

Will merge it.

PitNikola (approved):

crstauf:
> I'm afraid this kind of argument will not work well here.

That's [what was said](https://github.com/raycast/script-commands/pull/127#discussion_r532802022): "I would avoid brackets". Meaning that's your personal preference, not a requirement of script commands. If you meant to say it was a requirement for script commands, then my response would be different.

> seeing so much resistance in PR reviews makes it much harder to collaborate

I agree: never have I received so much criticism over minute items before. If this is hoped to be a public catalog of Script Commands contributed by a thriving community, then the criticism level needs to come way down, to allow script authors to be creative and express themselves. Several times I've been discouraged nearly enough to stop contributing and pull out of the community entirely, due to the strength of resistance encountered when sincerely asking questions or opening a  discussion. I hope that the Raycast team and others in the community can allow more creativity and personality from script contributors. A thriving Raycast community would be a wonderful sight to behold.

PR #140, opened 2020-12-01
MDN JS Docs

traynham:
## Description

This script command searches through the JavaScript portion of the MDN (Mozilla Developer Network) Web Docs site.

## Type of change

- [√] New script command

## Screenshot

mdn-js

This script requires Nodejs which can be installed from https://nodejs.org/en/download/

## Checklist

- [√] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)


commands/web-searches/mdn.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+
+// Dependency: This script requires Nodejs.
+// Install Node: https://nodejs.org/en/download/
+//
+// Required parameters:
+// @raycast.schemaVersion 1
+// @raycast.title MDN JS Docs
```
PitNikola, line 8:
```suggestion
// @raycast.title Search in MDN JS Docs
```

PitNikola (changes requested):
Thanks for contribution! Considering that this script just opens URL, we would recommend converting it to bash script to avoid NodeJS dependency (see [section about dependencies](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#scripts-that-require-installation-of-runtimes-and-dependencies) in our guidelines). This should improve portability.

See [Google Search example](https://github.com/raycast/script-commands/blob/master/commands/web-searches/google-search.sh) for how to do that.

traynham:
> Thanks for contribution! Considering that this script just opens URL, we would recommend converting it to bash script to avoid NodeJS dependency (see [section about dependencies](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#scripts-that-require-installation-of-runtimes-and-dependencies) in our guidelines). This should improve portability.
> 
> See [Google Search example](https://github.com/raycast/script-commands/blob/master/commands/web-searches/google-search.sh) for how to do that.

To be frank, the point of my submission is that it is written in Node and shows how to grab the argument and use exec to open the url. Converting it to a shell script is not difficult, but I was hoping to share this simple script as node to help others get started.

PitNikola:
@traynham, I see. While I like the intention to provide others with example of how to write node js scripts and get arguments, unfortunately I won't be able to merge it since it doesn't follow our Contribution guidelines. Perhaps you know a better example for a script that could leverage NodeJS runtime?

unnamedd:
@PitNikola what do you think if the community create some other template examples in different languages? I.e: NodeJS, Ruby, Python and others? 

I guess this might solve the problem brought by issue, right?

traynham:
I have been trying to think of a way to use features specific to Node, but truthfully you can do all the same things in each language. What excites me about Node is it is javascript so you if you know JS for the browser you pretty much know node. You could potentially have a larger group of people familiar with JS than shell scripting. Don't get me wrong, I've got dozens and dozens of shell scripts that I use -- so I'm not anti shell script. :)

So I am at a loss on how to move forward with this script here, as its purpose was to be very simple. I will likely just create a gist for it. Given the opportunity, I'd love to submit it as an example starting point for others to use. I did spend a few hours learning a few of the features needed to get it to work, and that could save others some time.

Thanks for taking the time to look at and consider my script.

unnamedd:
Hi @traynham,

I have a suggestion for you. What do you think to make a small conversion in your script and turn it into a template? If you check the [templates](https://github.com/raycast/script-commands/tree/master/templates) folder you will see two examples there: one for shell and one for swift. Would be great to have more examples there to help others. No one need to lose anything and contributing to a template for the next script command developer, I'm sure will be a great addition to this repository. Don't you think?

traynham:
@unnamedd - Sounds interesting, I'll take a look.

traynham:
@PitNikola, @unnamedd - would this script-command.template.js script be acceptable?

PitNikola:
Hey, @traynham, yes, I think it's great idea to add it as a template (thanks @unnamedd for suggesting).

The template looks good, thanks for preparing it! Are you planning to convert MDN web search command to bash? If not, then feel free to leave it out from this PR.

traynham:
@PitNikola - I went ahead and converted the script to a shell script. I also added an optional "topic" param, set the title accordingly, and added PNGs for light and dark mode. Hopefully that covers everything. Please let me know if I've missed anything.

commands/web-searches/mdn.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Search in MDN (Mozilla) Docs by topic
```
PitNikola, line 5:
```suggestion
# @raycast.title Search in MDN (Mozilla) Docs by Topic
```

commands/web-searches/mdn.sh
```diff
@@ -0,0 +1,20 @@
...
+# @raycast.mode silent
+#
+# Optional parameters:
+# @raycast.packageName Web Searches
+# @raycast.icon images/mdn_light.png
+# @raycast.iconDark images/mdn_dark.png
+# @raycast.argument1 { "type": "text", "placeholder": "js, css, html", "optional": true}
+# @raycast.argument2 { "type": "text", "placeholder": "query" }
```
PitNikola, line 13: What do you think about making "topic" second parameter since it's optional and query first? This way you would just need to press `Tab` once if you want to omit topic?

No strong feelings though, up to you

PitNikola (commented):
Looks good, just minor suggestions

commands/web-searches/mdn.sh
traynham: My personal preference is to keep the topic param first because it seems more hierarchal to me. I also think the topic param would be used more than not. In my testing, this seems the most natural route to me.

PR #141, opened 2020-12-02
WIP: feat(System): caffeinate

maximkrouk:
## Description

Prevents devices from sleeping

## Type of change

- [x] New script command

## Screenshot

<p align="center">
Screen Shot 2020-12-06 at 21 16 32

Screen Shot 2020-12-06 at 21 17 23

Screen Shot 2020-12-02 at 06 25 29

Screen Shot 2020-12-06 at 21 15 04

Screen Shot 2020-12-06 at 21 15 50

Screen Shot 2020-12-02 at 06 25 58

Screen Shot 2020-12-02 at 06 26 13

<h2>UPD:<h2>

Screen Shot 2020-12-06 at 21 08 56

Screen Shot 2020-12-06 at 21 09 16

Screen Shot 2020-12-06 at 21 10 13

</p>

None

## Checklist

### DIFF
--- a/commands/system/start-caffeinate.sh
+++ b/commands/system/start-caffeinate.sh
@@ -1,0 +1,22 @@
+#!/bin/zsh
+
+# @raycast.schemaVersion 1
+# @raycast.title Caffeinate Start
+# @raycast.mode silent
+# @raycast.author Maxim Krouk
+#
+# @raycast.authorURL https://github.com/maximkrouk
+# @raycast.description Prevents your device from sleepping
+# @raycast.icon ☕️
+# @raycast.packageName System
+# @raycast.argument1 { "type": "text", "placeholder": "Seconds" }
+
+value=$1
+regex='^[0-9]+$'
+if ! [[ $value =~ $regex ]] ; then
+  echo "Provide a number, please  ❌"
+else
+  caffeinate -u -t $1 '&'
+  echo "Won't sleep for" $value "seconds"
+fi
+
--- a/commands/system/start-caffeinate.sh
+++ b/commands/system/start-caffeinate.sh
@@ -9,14 +9,18 @@
 # @raycast.description Prevents your device from sleepping
 # @raycast.icon ☕️
 # @raycast.packageName System
-# @raycast.argument1 { "type": "text", "placeholder": "Seconds" }
+# @raycast.argument1 { "type": "text", "placeholder": "Seconds", "optional": true }
 
-value=$1
-regex='^[0-9]+$'
-if ! [[ $value =~ $regex ]] ; then
-  echo "Provide a number, please  ❌"
-else
-  caffeinate -u -t $1 &
-  echo "Won't sleep for" $value "seconds"
+if [ -z $1 ]
+then
+    caffeinate -u &
+    echo "Won't sleep forever" && exit 0
 fi
 
+regex='^[0-9]+$'
+if ! [[ $1 =~ $regex ]] ; then
+    echo "Provide a number, please" && exit 1
+else
+    caffeinate -u -t $1 &
+    echo "Won't sleep for" $1 "seconds" && exit 0
+fi
--- a/commands/system/start-caffeinate.sh
+++ b/commands/system/start-caffeinate.sh
@@ -16,7 +16,7 @@ regex='^[0-9]+$'
 if ! [[ $value =~ $regex ]] ; then
   echo "Provide a number, please  ❌"
 else
-  caffeinate -u -t $1 '&'
+  caffeinate -u -t $1 &
   echo "Won't sleep for" $value "seconds"
 fi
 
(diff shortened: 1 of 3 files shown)


maximkrouk:
I also have a draft for tracking these processes

```bash
#!/bin/zsh

# @raycast.schemaVersion 1
# @raycast.title Caffeinate
# @raycast.mode inline
# @raycast.author Maxim Krouk
#
# @raycast.authorURL https://github.com/maximkrouk
# @raycast.description Displays active caffeinates.
# @raycast.icon ☕️
# @raycast.packageName System
# @raycast.refreshTime 15s

result=$(ps aux | grep caffeinate | awk '{print $9 " " $14}') # Output example - "5:32PM 1231"

# todo : calculate time
regex='^[0-9]+$'
if ! [[ $result =~ $regex ]] ; then
  echo "No caffeine here"
else
  echo "Won't sleep for" $result "seconds"
fi
```

commands/system/start-caffeinate.sh
```diff
@@ -0,0 +1,22 @@
...
+# @raycast.authorURL https://github.com/maximkrouk
+# @raycast.description Prevents your device from sleepping
+# @raycast.icon ☕️
+# @raycast.packageName System
+# @raycast.argument1 { "type": "text", "placeholder": "Seconds" }
+
+value=$1
+regex='^[0-9]+$'
+if ! [[ $value =~ $regex ]] ; then
+  echo "Provide a number, please  ❌"
```
PitNikola, line 17: @maximkrouk, in latest version (1.3.0) we improved error reporting in silent and compact mode. Now if you exit with non-zero status, we'll display warning emoji with last output line. Like this:

CleanShot 2020-12-03 at 12 21 45@2x

Shall we leverage it here? Something like:
```
echo "Provide a number please"
exit 1
```

PitNikola (changes requested):
Really handy one! Btw, thanks for providing such nice screenshots, love it!

I have only minor suggestion (due to our recent improvement in 1.3.0).

commands/system/start-caffeinate.sh
maximkrouk: Ok, I'll fix it soon

maximkrouk: Btw ty 4 the compliment, maybe it worth adding cmd+5 screenshots to contribution guidelines 😅

unnamedd: > @maximkrouk, in latest version (1.3.0) we improved error reporting in silent and compact mode. Now if you exit with non-zero status, we'll display warning emoji with last output line. Like this:
> 
> CleanShot 2020-12-03 at 12 21 45@2x
> 
> Shall we leverage it here? Something like:
> 
> ```
> echo "Provide a number please"
> exit 1
> ```

@PitNikola it will be nice if you move this comment/documentation to the README. What do you think?

maximkrouk:
Also I'd like to implement custom output for longest-caffeinate command, I don't know how to implement it in bash, may I do it in Swift?

I can get a sorted array of caffeinate start times and optional numbers of seconds with
```bash
#input
ps aux | grep '\d caffeinate' | sort -k14 | awk '{print $9 " " $14}'

#output
4:48PM 
4:48PM 
4:48PM 
4:56PM 1000
4:57PM 1100
```
I need to detect if there is an output without a timeout parameter and get amount of seconds from midnight plus duration for each parameter in this array.

Also I think it worth opening another PR for that command if u still want to see it implemented in bash

commands/system/start-caffeinate.sh
PitNikola: @unnamedd, good point. Added to Error Handling section in README

maximkrouk: Fixed error handling, made an argument optional, but now the script doesn't work as expected 😢
Note:
- `&` operator does not work in `caffeinate -u -t $1 &` expression, so the command blocks the execution and the message doesn't show up.
- Also I added a basic script for the longest active caffeinate command
- I updated screenshots 👆🙂

PitNikola: > & operator does not work in caffeinate -u -t $1 & expression, so the command blocks the execution and the message doesn't show up (but it works fine in terminal).

I'm gonna take a look into it.

PitNikola:
I can confirm that the process doesn't exit when using `&`, which makes this particular script a bit tricky. I'm gonna need to put this PR on hold until we find a way to deal with this. Sorry for inconvenience.

thomaspaulmann:
I'm gonna close this PR for now since we don't support it properly. We can pick it up as soon as we have support for async execution. This way we keep the PR review pipeline clean.

PitNikola:
@maximkrouk, it seems like there is a way to resolve the async execution issue by adding `&>/dev/null` when executing caffeinate as was discovered in [this PR](https://github.com/raycast/script-commands/pull/481) 🎉

PR #148, opened 2020-12-05
Set scripts and script templates executable

crstauf:
## Description

Set script and script templates executable. Possibly not much need for templates to be executable, but might as well set them, right?

## Type of change

- [x] Other (Specify)

## Checklist


PitNikola:
Yeah, good catch, thanks for updating them 👍

PR #158, opened 2020-12-09
add url decode/encode and escape-shell scripts

daern91:
## Description

PR adds three new scripts, one for decoding URLs, one for encoding URLs, and one script for escaping shell characters. The last one is especially handy for me when copying paths from remote servers and wanting to paste them with escaped backspaces.

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

The first two should be self-explanatory, so I'm only attaching the last one. Please let me know if you want gifs for the first two as well.

Escaping shell
CleanShot 2020-12-09 at 17 04 02

## Checklist

### DIFF
--- a/commands/developer-utils/encode-url.sh
+++ b/commands/developer-utils/encode-url.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Encode URL
+# @raycast.mode silent
+# @raycast.packageName Developer Utilities
+
+# Optional parameters:
+# @raycast.icon 💻
+
+# Documentation:
+# @raycast.description Encodes clipboard content url and copies it again.
+
+pbpaste | ( curl -Gso /dev/null -w %{url_effective} --data-urlencode @- "" | cut -c 3- || true) | pbcopy
+echo "Encoded URL"
--- a/commands/developer-utils/escape-shell-chars.sh
+++ b/commands/developer-utils/escape-shell-chars.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Escape shell string
+# @raycast.mode silent
+# @raycast.packageName Developer Utilities
+
+# Optional parameters:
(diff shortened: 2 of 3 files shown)


commands/developer-utils/escape-shell-chars.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Escape shell string
```
thomaspaulmann, line 5: How about a title like this:
```suggestion
# @raycast.title Escape String for Shell
```

daern91: Sounds great to me. Any preference for the file name as well?

commands/developer-utils/encode-url.sh
```diff
@@ -0,0 +1,16 @@
...
+# @raycast.mode silent
+# @raycast.packageName Developer Utilities
+
+# Optional parameters:
+# @raycast.icon 💻
+
+# Documentation:
+# @raycast.description Encodes clipboard content url and copies it again.
+
+pbpaste | ( curl -Gso /dev/null -w %{url_effective} --data-urlencode @- "" | cut -c 3- || true) | pbcopy
```
thomaspaulmann, line 15: Cool idea to use `cURL` for encoding the URL.

commands/developer-utils/escape-shell-chars.sh
```diff
@@ -0,0 +1,16 @@
...
+# @raycast.packageName Developer Utilities
+
+# Optional parameters:
+# @raycast.icon 💻
+
+# Documentation:
+# @raycast.description Escapes shell character string and copies it again.
+
+escaped=$(pbpaste) && printf '%q' "$escaped" | pbcopy
+echo "Escaped shell string"
```
thomaspaulmann, line 16:
```suggestion
echo "Escaped string for shell"
```

thomaspaulmann (approved):
Yeah! Thanks for the addition @daern91. I'm pretty sure the commands will be useful for many devs.

PR #164, opened 2020-12-10
Add Script Command: play-movie

crstauf:
## Description

Play specified movie in TV.

## Type of change

- [x] New script command

## Screenshot

Screen Shot 2020-12-10 at 4 40 32 PM

## Checklist

### DIFF
--- a/commands/media/apple-tv/apple-tv-play-movie.applescript
+++ b/commands/media/apple-tv/apple-tv-play-movie.applescript
@@ -1,0 +1,21 @@
+#!/usr/bin/osascript
+
+## Contributions welcome to also play TV shows.
+
+# @raycast.title Play Movie
+# @raycast.author Caleb Stauffer
+# @raycast.authorURL https://github.com/crstauf
+# @raycast.description Play specified movie from library in TV.
+
+# @raycast.icon images/apple-tv-logo.png
+# @raycast.mode silent
+# @raycast.packageName TV
+# @raycast.schemaVersion 1
+
+# @raycast.argument1 { "type": "text", "placeholder": "Movie" }
+
+on run argv
+	tell application "TV"
+		to play track (item 1 of argv)
+	end tell
+end run
--- a/commands/media/apple-tv/apple-tv-play-movie.applescript
+++ b/commands/media/apple-tv/apple-tv-play-movie.applescript
@@ -16,6 +16,11 @@
 
 on run argv
 	tell application "TV"
-		play track (item 1 of argv)
+		try
+			activate
+			play track (item 1 of argv)
+		on error
+			log "Unable to play \"" & (item 1 of argv) & "\""
+		end try	
 	end tell
 end run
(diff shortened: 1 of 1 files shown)


commands/media/apple-tv/apple-tv-play-movie.applescript
```diff
@@ -0,0 +1,21 @@
...
+# @raycast.packageName TV
+# @raycast.schemaVersion 1
+
+# @raycast.argument1 { "type": "text", "placeholder": "Movie" }
+
+on run argv
+	tell application "TV"
+		to play track (item 1 of argv)
+	end tell
+end run
```
thomaspaulmann, line 21: This script doesn't seem to work. Changing it to the following makes it run but I get the following execution error: `TV got an error: Can’t make some data into the expected type. (-1700)`

```suggestion
on run argv
  tell application "TV"
    play track (item 1 of argv)
  end tell
end run
```

commands/media/apple-tv/apple-tv-play-movie.applescript
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/osascript
+
+## Contributions welcome to also play TV shows.
```
thomaspaulmann, line 3: Lets remove this. Not sure if that is really needed.

thomaspaulmann (changes requested):
I tried the AppleScript but it didn't work for me. I fixed the syntax error but get an execution error that I can't decrypt. Any idea how to make it work?

commands/media/apple-tv/apple-tv-play-movie.applescript
crstauf: Not needed, but wanted.

crstauf:
I'll double check that it matches my local version, which I know works. Don't know why they wouldn't match, but I'll double check.

crstauf:
@thomaspaulmann The movie name you typed in... was it one word or multiple?

crstauf:
Hmm, after playing with it, I can't get it back to working. 😕 

If I can't stabilize it within the next week I'll close.

commands/media/apple-tv/apple-tv-play-movie.applescript
crstauf: @thomaspaulmann I just tried every movie in my library, and every single one worked:
- single word titles
- multi-word titles
- titles with special characters

(I was honestly surprised when `Mission: Impossible - Ghost Protocol` started playing.)

Can you please provide the movie titles you entered that failed?

thomaspaulmann: I entered `pulp fiction` and it returned the error. Might be worth to do a try-catch-block around the play method and give a nicer error message, e.g. "Move not found" and exit with status code 1. 

This would give the user a better result. Also, I don't use Apple TV heavily :-)

thomaspaulmann:
> @thomaspaulmann The movie name you typed in... was it one word or multiple?

Added a comment on our discussion.

thomaspaulmann:
Thanks for adding the try-catch-block. It's a much nicer user-facing error message that the underlying error provides.

PR #171, opened 2020-12-29
New script command: What is my AWS region

c1982:
## Description

This command provide finds which AWS region the IP address is in. it's not required AWS credential. 

usage: copy ipv4 address to clipboard and run script over raycast.

output:
---
```json
{
  "ip_prefix": "65.0.0.0/14",
  "region": "ap-south-1",
  "service": "EC2",
  "network_border_group": "ap-south-1",
  "ip": "65.0.132.249"
}
```

## Type of change

- [x] New script command

* https://ip-ranges.amazonaws.com/ip-ranges.json
* curl (https://curl.se/)
* jq (https://stedolan.github.io/jq/)
* grepcidr (http://www.pc-tools.net/unix/grepcidr/)

## Checklist


PitNikola (changes requested):
Thanks for your contribution! Generally looks good, just minor things:

1. The screenshot doesn't need to be committed, we need it only for PR review. Let's remove it.
2. What do you think about using input argument instead of clipboard history? It would make it more explicit. Check out [this section](https://github.com/raycast/script-commands#passing-arguments) on how to pass arguments. Not a hard requirement, so leaving it up to you. If you decide to leave clipboard history approach, I would suggest updating error message to something like "Couldn't read valid IPv4 address from clipboard" to hint that it's reading it from there.

c1982:
Hi @PitNikola thanks for review.

Yes, input argument is useful and more explicit a most cases. I think clipboard better than input argument for this case. Input arguments add more one step for access to result.

PitNikola:
Makes sense, thanks!

PR #172, opened 2021-01-04
Add command to check if a website is up or not

jesse-c:
## Description

Add command to check if a website is up or not. It uses the isitup.org service.

## Type of change

- [x] New script command

## Screenshot

image

N/A

## Checklist

### DIFF
--- a/commands/developer-utils/is-it-up.sh
+++ b/commands/developer-utils/is-it-up.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# Dependency: requires jq (https://stedolan.github.io/jq/)
+# Install via Homebrew: `brew install jq`
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Is It Up?
+# @raycast.mode compact
+# @raycast.packageName Developer Utils
+#
+# Optional parameters:
+# @raycast.icon 🩺
+# @raycast.argument1 { "type": "text", "placeholder": "Website" }
+#
+# Documentation:
+# @raycast.description Check if a website is up
+# @raycast.author Jesse Claven
+# @raycast.authorURL https://github.com/jesse-c
+
+if ! command -v jq &> /dev/null; then
+  echo "jq is required (https://stedolan.github.io/jq/).";
+  exit 1;
+fi
+
+status_code=$(curl --silent "https://isitup.org/$1.json" | jq '.status_code')
+
+# Sample output:
+#
+# {
+#   "domain": "duckduckgo.com",
+#   "port": 80,
+#   "status_code": 1,
+#   "response_ip": "52.142.124.215",
+#   "response_code": 200,
+#   "response_time": 0.021
+# }
+
+case $status_code in
+  1) echo "Up: $1"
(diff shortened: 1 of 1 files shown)


commands/developer-utils/is-it-up.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
```
thomaspaulmann, line 2: Let's add a hint that this script uses a dependency as described in the contribution guidelines even though `jq` is a rather common one.
```suggestion

# Dependency: This script requires `jq` cli installed: https://stedolan.github.io/jq/
# Install via homebrew: `brew install jq`

```

commands/developer-utils/is-it-up.sh
```diff
@@ -0,0 +1,46 @@
...
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Is It Up?
+# @raycast.mode compact
```
thomaspaulmann, line 6: Could also be neat to have as an `inline` script for specific websites that you want to keep observing.

commands/developer-utils/is-it-up.sh
```diff
@@ -0,0 +1,46 @@
...
+# @raycast.packageName Developer Utils
+#
+# Optional parameters:
+# @raycast.icon 🩺
+# @raycast.currentDirectoryPath ~
+# @raycast.needsConfirmation false
```
thomaspaulmann, line 12: The default value is `false`. You can remove the parameter.

```suggestion
```

commands/developer-utils/is-it-up.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Is It Up?
+# @raycast.mode compact
+# @raycast.packageName Developer Utils
+#
+# Optional parameters:
+# @raycast.icon 🩺
+# @raycast.currentDirectoryPath ~
```
thomaspaulmann, line 11: Is this necessary? I assume that the script works from every directory?

thomaspaulmann (changes requested):
Script looks good. I've added some inline comments, nothing major. Most importantly, is it necessary to execute the script from `~`? This could be different on user machines but I also don't see anything that would prevent somebody to execute it from somewhere else. If not, I would remove the `currentWorkingDirectory` parameter.

commands/developer-utils/is-it-up.sh
jesse-c: Done.

commands/developer-utils/is-it-up.sh
jesse-c: It would! I've removed it.

commands/developer-utils/is-it-up.sh
jesse-c: It would! I guess people could copy the script and remove the argument?

thomaspaulmann: Yeah, let them be creative 🎨

PR #174, opened 2021-01-05
Slack: set Spotify song as slack status

alongat:
## Description

1. First script is a ported script I had that set Slack status to the current song playing on Spotify
2. Second sets the status_emoji to either :office: (WFO) or :house_with_garden: (WFH) according to the WiFi name.

## Type of change

- [X] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

Kapture 2021-01-05 at 22 03 54

## Checklist


alongat:
Looks like I needed to use `inline` and not `compact` for refreshTime to work (so the gif is showing the `compact` mode)

thomaspaulmann:
Thanks for the script command. That's quite an advanced one! Cool idea to let your co-workers know about your music taste :-) I was wondering if it makes sense to split the command into two. One for the status update of the song and the other for the Wifi handling. The separate commands seem more applicable to more users. What do you think?

alongat:
@thomaspaulmann Sure! sounds reasonable :) will do.

thomaspaulmann:
Cool! Looking forward to the split commands.

thomaspaulmann (changes requested):
As discussed, let's split this command into multiple ones to make it more suitable for wider adoption.

alongat:
@thomaspaulmann updated as discussed 💪 indeed much cleaner

commands/communication/slack/set-slack-status-spotify.sh
```diff
@@ -0,0 +1,118 @@
...
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Set Status in Slack to Spotify Song
+# @raycast.mode inline
+# @raycast.refreshTime 30s
+
+# Optional parameters:
+# @raycast.packageName Slack
+# @raycast.icon ../../media/images/spotify-logo.png
```
thomaspaulmann, line 15: Tricky. I was wondering if we should rather take the Slack logo here since it changes the Slack status. Then you can also reference it easier. If not, you might want to copy the Spotify logo closer to the script to make sure that it's easy for others to copy.

commands/communication/slack/set-slack-wfh-status.sh
```diff
@@ -0,0 +1,75 @@
...
+
+# Parameters
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Set WFH Status in Slack
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.packageName Slack
```
thomaspaulmann, line 13: Let's set an icon to make sure that the command looks nice :-) Since this is Slack, I would use the Slack logo. Alternatively you can just use an emoji? Maybe 🏡?

thomaspaulmann (changes requested):
Thanks for splitting. Much better now that the functionality is decoupled. I left some minor comments to make the commands stand out. I would suggest adding a Slack logo to the repo that you can reference.

commands/communication/slack/set-slack-wfh-status.sh
alongat: added slack logo on both 💪 
(sorry for the long response cycle too much work during the week 😅 )

thomaspaulmann: All good. Happy that we got it sorted.

thomaspaulmann (approved):

PR #177, opened 2021-01-07
feat: add Wi-Fi and Bluetooth toggle script

vincentdoerig:
## Description

Adds two system scripts, one that turns your wifi on/off based on 

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

Wi-Fi:
image
Bluetooth:
image

## Checklist


commands/system/bluetooth.applescript
```diff
@@ -0,0 +1,25 @@
...
+
+# @raycast.icon images/bluetooth.png
+# @raycast.mode silent
+# @raycast.packageName System
+# @raycast.schemaVersion 1
+
+tell application "System Events"
+  tell process "ControlCenter"
+    click menu bar item "Bluetooth" of menu bar 1
+    set BluetoothSwitch to checkbox "Bluetooth" of group 1 of window "Control Centre"
```
thomaspaulmann, line 16: Mmh... Looks like this line makes trouble on macOS Big Sur:

```
execution error: System Events got an error: Can’t get group 1 of window "Control Centre" of process "ControlCenter". Invalid index. (-1719)
```

Any idea on how to fix this?

thomaspaulmann: There is also the possibility to reverse-engineer how [blueutil](https://github.com/toy/blueutil/blob/master/blueutil.m) works and write a short Swift script that toggles Bluetooth.

commands/system/wifi.applescript
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/osascript
+
+# @raycast.title Wi-Fi
```
thomaspaulmann, line 3: Let's be consistent with the naming of the command for Bluetooth and others in this repo.
```suggestion
# @raycast.title Toggle Wi-Fi
```

commands/system/wifi.applescript
```diff
@@ -0,0 +1,25 @@
...
+# @raycast.author Vincent Dörig
+# @raycast.authorURL https://github.com/vincentdoerig
+# @raycast.description Turn your Wi-Fi on or off.
+
+# @raycast.icon images/wifi-dark.png
+# @raycast.iconDark images/wifi.png
+# @raycast.mode silent
+# @raycast.packageName System
+# @raycast.schemaVersion 1
+# @raycast.argument1 { "type": "text", "placeholder": "on/off" }
```
thomaspaulmann, line 13: I would drop this input and make it toggle, similar to the Bluetooth one. It's usually simpler to execute it.

commands/system/wifi.applescript
```diff
@@ -0,0 +1,25 @@
...
+
+on run argv
+  if item 1 of argv is "off" then
+    do shell script "networksetup -setairportpower en1 off"
```
thomaspaulmann, line 17: Is it always `en1`? If not, I would recommend extracting this as a property and add a comment how to find out what to put there. This way it's easy for others to adjust the script to their local setup.

thomaspaulmann (changes requested):
Hey @vincentdoerig, thanks a lot for the contribution. Especially the bluetooth command will make @pacocoursey happy :-) Unfortunately, the command doesn't work on macOS Big Sur. Any chance you could fix it to make sure that it works for users on the latest OS? If there is something to manually adjust, e.g. naming of some processes, I recommend to extract those to properties and add a comment for users to simply adopt it to their local setup.

I added some inline comments. Could you also re-name the script files to `toggle-bluetooth` and `toggle-wifi`? It makes clearer what the commands do and leave room for other wireless connection related commands.

Thanks again for this great contribution! 🎉

commands/system/bluetooth.applescript
unnamedd: > Mmh... Looks like this line makes trouble on macOS Big Sur:
> 
> ```
> execution error: System Events got an error: Can’t get group 1 of window "Control Centre" of process "ControlCenter". Invalid index. (-1719)
> ```
> 
> Any idea on how to fix this?

There's a problem in this kind of approach and I already discussed it with Petr. It is related to the language. You're using the UK grammar (Control Centre) and Valentin configured the text using the US grammar (Control Center). Every user who decides to use those scripts will have to "translate" the strings to their own language, i.e: spanish, portuguese, italian and so on. The invalid index confirms that to me.

commands/system/wifi.applescript
unnamedd: Not always Thomas. In my laptop, the WiFi is configured on `en0`.

commands/system/bluetooth.applescript
vincentdoerig: yes, @unnamedd is correct. The error is because of the different language localisations (I'm using british english, hence the _centre_). What's the best approach to inform the user to edit the script according to their language?

commands/system/wifi.applescript
vincentdoerig: Right, I changed it to toggle, however I imagine we will run into localisation problems again, since we check if the response from `networksetup -getairportnetwork [interface]` contains the word "off".

thomaspaulmann: Not sure if the CLI tool `networksetup` is localized. I would expect that it always returns on/off.

vincentdoerig: Alright, great! Then it shouldn't be a problem.

commands/system/bluetooth.applescript
thomaspaulmann: Mmh... If we can't make it language-agnostic, I would vouch for extracting the strings as constants and add a comment that explains on how to get the names.

vincentdoerig: I can't think of a command or something else that would return the names in a specific language. Any ideas?

thomaspaulmann: Me neither. I think this script just needs a custom setup depending on the users locale. For this kind of script commands, we name them with `template` in the filename to make sure that folks need to configure them, e.g. `toggle-wifi.template.applescript`. Raycast doesn't show scripts that include `template` in the filename.

As mentioned, let's just extract the language-dependent bits of the script into constants at the top of the script and add a comment there that describes on how to find out what to put there.

commands/system/wifi.applescript
```diff
@@ -1,25 +1,25 @@
...
-    do shell script "networksetup -setairportpower en1 off"
-    do shell script "echo Wi-Fi turned off"
-  else if item 1 of argv is "on" then
-    do shell script "networksetup -setairportpower en1 on"
-    do shell script "echo Wi-Fi turned on"
-  else
-    do shell script "echo Invalid param provided"
-  end if
-end run
+# Wi-Fi interface, should normally be either `en0` or `en1`
```
thomaspaulmann, line 14: How do you find out if it's `en0` or `en1` or something else? Opening the `Network` preferences?

vincentdoerig: option click on the wifi icon in the menu bar shows it 
image

commands/system/bluetooth.applescript
vincentdoerig: 👍 will do in the next couple of days or so

commands/system/wifi.applescript
unnamedd: The other way is running `ifconfig` on terminal and checking which interface has any IP (inet) defined.

unnamedd: Example: 
image
image

thomaspaulmann: Ok cool. Let's add one of those ways to check the name of the interface to the comment.

thomaspaulmann (approved):
Thanks for adding the comments and doing all the changes. Let's get this change landed.

PR #180, opened 2021-01-11
Add Python and Ruby templates

Matt-Gleich:
## Description

Add script templates for the ruby and python programming languages.

## Type of change

- [ ] New script command
- [ ] Bug fix
- [x] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

N/A

python3 and ruby installed

## Checklist

### DIFF
--- a/templates/script-command.template.py
+++ b/templates/script-command.template.py
@@ -1,4 +1,4 @@
-#!/usr/local/bin/python3
+#!/usr/bin/env python3
 
 # Raycast Script Command Template
 #
--- a/templates/script-command.template.py
+++ b/templates/script-command.template.py
@@ -2,6 +2,9 @@
 
 # Raycast Script Command Template
 #
+# Dependency: This script requires Python 3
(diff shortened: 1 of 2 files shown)


templates/script-command.template.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/local/bin/python3
```
fe9lix, line 1: I'd suggest that as shebang we use `#!/usr/bin/env python3`, which is a bit more portable.

templates/script-command.template.rb
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/ruby
```
fe9lix, line 1: Same as for Python above: let's use `#!/usr/bin/env ruby`

fe9lix (changes requested):
Review done.

fe9lix (approved):

PR #186, opened 2021-01-16
Pixelmator Pro: add applescript to open image from clipboard

bdcarr:
## Description

This adds a new command which uses applescript to grab an image from your clipboard and open it in a new document in the image editing app, Pixelmator Pro. There's a bit of inconsistency in whether it opens in a new window or a new tab, but I've opened a ticket with Pixelmator and hopefully it won't require changes to the script.

## Type of change

- [X] New script command

## Screenshot

https://user-images.githubusercontent.com/1940488/104797008-006f9b00-580f-11eb-88d1-aa24a2faa171.mp4

Technically it depends on having Pixelmator Pro installed, but I figure you won't be installing this if you don't already have it. I've put a note in the script comments anyway.

## Checklist


commands/apps/pixelmator-pro/pixelmator-open-clipboard.applescript
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/osascript
```
unnamedd, line 1: Hi @bdcarr, as we already have some scripts doing similar work, the best folder to place this Script Command is `navigation`.

unnamedd (changes requested):
Nice script command! Should help a lot those who use Pixelmator.

ps: because of this script command I decided to install and try it. Thanks! :D

bdcarr:
Thanks @unnamedd! It's the best image editor I've used, though I pretty much only make crap memes with it.

* Moved script and image to `navigation` folder
* Renamed script to fit better with the other scripts (I think)
* Added an explicit `return` at the end to stop the document ID popping up

bdcarr:
Is it okay to leave the package name as Pixelmator Pro? From a UI perspective, I think it helps communicate what the command does without making the title too long.

If the package name has to be Navigation though, I'll change the title.

unnamedd:
> Is it okay to leave the package name as Pixelmator Pro? From a UI perspective, I think it helps communicate what the command does without making the title too long.
> 
> If the package name has to be Navigation though, I'll change the title.

Not only okay, but it's also recommended to be like that. In that way, you will only write in the search "pixel" and your Script Command will pop up in the results. This is the best way to group some script commands under the same packageName.

Take a look on the [developer-utils](https://github.com/raycast/script-commands/tree/master/commands/developer-utils) folder! ;)

unnamedd (approved):

PitNikola:
Cool one! Thanks for your contribution!

PR #20, opened 2020-10-31
Add Script Commands: URL IPv4/IPv6

crstauf:

PR #196, opened 2021-01-18
Domain to IP

pernielsentikaer:
## Description

Lookup the IP address of any given domain

## Screenshot

3uu0amajoq0

## Checklist


unnamedd:
Hey @pernielsentikaer, very interesting first contribution!

unnamedd (approved):

pernielsentikaer:
Seems to be doing the same as #20 which also copies the IP (got confused by the name) - I think this script should be removed again, or maybe incorporate so the other script in #20 would have a way to not copy to clipboard, I'm not sure if thats possible yet?

PitNikola:
I would double check with @crstauf if he's up for those IP scripts updated to input mode. Though I can imagine that in some cases using clipboard might be a better flow for some people. I'm fine if we have both in this repository.
If we end up adding this one, let's move it also to `ip` sub-directory.

thomaspaulmann (changes requested):
Command looks good. Let's move it to the `ip` subfolder and then it should be good to go into master. Great first contribution 👏

crstauf:
This is a duplicate of two commands already in the repo: [get-url-ip-v4.sh](https://github.com/raycast/script-commands/blob/master/commands/developer-utils/ip/get-url-ip-v4.sh) and [get-url-ip-v6.sh](https://github.com/raycast/script-commands/blob/master/commands/developer-utils/ip/get-url-ip-v6.sh), unless I'm missing something...?

thomaspaulmann:
Oh good catch! I was under the impression that it still uses the clipboard as input instead of arguments. Gonna close this then because it's a duplicate.

PR #198, opened 2021-01-18
Add a script to visit a specific GitHub repo, pull request, or issue

PSalant726:
## Description

Do you already know which GitHub repo you want to open, and don't need to use the GitHub extension to search for a repository? Better yet, do you already know the specific pull request and/or issue by number? Then use this script command!

## Type of change

- [x] New script command

### Visit a repo directly
Screen Shot 2021-01-23 at 6 09 30 PM

### Visit a PR directly
Screen Shot 2021-01-23 at 6 09 52 PM

### Visit an Issue directly
Screen Shot 2021-01-23 at 6 10 02 PM

None

## Checklist

### DIFF
--- a/commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
+++ b/commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
@@ -3,15 +3,14 @@
 # Required parameters:
 # @raycast.schemaVersion 1
 # @raycast.title Visit a Repository, Pull Request, or Issue
-# @raycast.mode compact
+# @raycast.mode silent
 #
 # Optional parameters:
 # @raycast.packageName GitHub
 # @raycast.icon images/github-logo.png
 # @raycast.iconDark images/github-logo-iconDark.png
 # @raycast.argument1 { "type": "text", "placeholder": "Organization/Repository" }
-# @raycast.argument2 { "type": "text", "placeholder": "Pull Request #", "optional": true }
-# @raycast.argument3 { "type": "text", "placeholder": "Issue #", "optional": true }
+# @raycast.argument2 { "type": "text", "placeholder": "Pull Request or Issue #", "optional": true }
 #
 # Documentation
 # @raycast.author Phil Salant
@@ -19,15 +18,8 @@
 # @raycast.description Open a repository, pull request, or issue on GitHub
 
 path=""
-if [[ -n "$2" && -n "$3" ]]; then
-  echo "Do not pass both a pull request and an issue."
-  exit 1
-elif [[ -n "$2" ]]; then
+if [ -n "$2" ]; then
   path="/pull/$2"
-elif [[ -n "$3" ]]; then
-  path="/issues/$3"
 fi
 
-echo "Opening $1$path on GitHub..."
 open "https://github.com/$1$path"
-exit 0
--- a/commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
+++ b/commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title GitHub Repository PRs & Issues
+# @raycast.mode compact
+#
+# Optional parameters:
+# @raycast.packageName GitHub
+# @raycast.icon images/github-logo.png
+# @raycast.iconDark images/github-logo-iconDark.png
+# @raycast.argument1 { "type": "text", "placeholder": "Organization/Repository" }
+# @raycast.argument2 { "type": "text", "placeholder": "Pull Request #", "optional": true }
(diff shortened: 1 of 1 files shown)


commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title GitHub Repository PRs & Issues
```
unnamedd, line 5: Phil, as Raycast uses the packageName to help in the search, you don't need to add `GitHub` here, then I suggest this: 
```suggestion
# @raycast.title Pull Requests & Issues
```

What do you think?

commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
```diff
@@ -0,0 +1,33 @@
...
+# @raycast.argument3 { "type": "text", "placeholder": "Issue #", "optional": true }
+#
+# Documentation
+# @raycast.author Phil Salant
+# @raycast.authorURL https://github.com/PSalant726
+# @raycast.description Open a repository, pull request, or issue on GitHub
+
+path=""
+if [[ -n "$2" && -n "$3" ]]; then
+  echo "Do not pass both a pull request and an issue."
```
unnamedd, line 23: Reading that instruction afterwards, I started to think that might be good if you split this script command in two, one to open Pull Requests and one another to open Issues. 

Having both options in the same script command, might lead the user to a confusion and as Script Commands must be simple, like a function, split in two may solve this problem, don't you think?

unnamedd (changes requested):

commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
PSalant726: Yea, I can remove `GitHub` but I would like to leave `Repository`. Both the PR and Issue arguments are optional, and the real benefit of the script is that a repository's root page can be opened when they aren't passed. Maybe something like `Visit a Repository, PR, or Issue`? I'm intentionally saying "visit" instead of "open" so as not to imply that a new repo, issue, or PR can be created using this script.

commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
PSalant726: The PR and Issue options are just nice-to-have features that were easy to tack on to the core functionality. For me, the real draw of this script is to be able to directly open a repository without first having to search, and so I prefer to keep everything combined.

To break them up would mean having all of the following:
1. A script that opens a repository, and has no additional functionality
1. A script that requires both an `organization/repository` string **and** an issue number be passed, and opens the issue
1. A script that requires both an `organization/repository` string **and** a PR number be passed, and opens the PR

I can refactor things accordingly, but IMO the latter two would rarely be used. I don't think a user would typically know these numbers (unless they _happen_ to know them, which is why I like having the feature at all).

It feels like a lot of bloat just to avoid a situation that I think is already easily avoidable, but I'm happy to make the change if that's the way you'd prefer to have it.

subtlepseudonym: I believe it's worth keeping the pull-request / issue functionality in this script.

However, I'm not a fan of having two mutually-exclusive, optional arguments. We could alleviate this by combining them into a single argument. The address `github.com/repo/pull/$ISSUE_NUM` will take you to both pull requests and issues, so there should be relatively few code changes.

commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
unnamedd: Make completely sense keep the `Repository`, I don’t know why I suggested that, after all, the pull requests and the issues belong to a repository! haha

commands/developer-utils/github/open-gh-repo-pr-or-issue.sh
unnamedd: @PSalant726 I don’t think will be needed to split the script command in two anymore. Even if I new, I completely forgot about the address which serve both PRs and Issue and fortunately, @subtlepseudonym just came here to bring that amazing information for us.

I don’t like either mutually-exclusive arguments, they tend to lead us to confusion. 

What we can have here, is a simplification of the code and in the argument, instead of two, we can have only one and in the placeholder, set something like “PR # or Issue #”, it will work for both anyway, doesn’t mean that it needs to be complex. What do you think?

PSalant726:
Without the need for error handling, this script now executes in `silent` mode.

unnamedd:
> Without the need for error handling, this script now executes in `silent` mode.

Great changes! Thanks a lot! Approved!

unnamedd (approved):

PR #211, opened 2021-01-28
Fix decimal date calculation

achirkof:
## Description

If day of the year returns number with prepended 0(zero) script can't
calculate progress with error: "Value too great for base" because
bash assumes it is octal number. 

## Type of change
- [x] Bug fix

## Checklist


thomaspaulmann:
Just got this error today as well :D Thanks for the fix @achirkof

PR #214, opened 2021-02-01
Add a Bitwarden package

PSalant726:
## Description

Adds a package of script commands that wrap the [Bitwarden CLI](https://bitwarden.com/help/article/cli/), and enable basic authentication, locking/unlocking, and searching of a vault.

The package has been added to a new `password-managers` directory, because none of the existing directories felt appropriate.

## Type of change

- [x] New [package of] script command[s]
- [x] Documentation ~update~ addition

## Screenshot

See [the included `README.md` file](https://github.com/PSalant726/script-commands/blob/0a29f0e05f1fd5a61ea3ef5ef6d924f9f278d68f/commands/password-managers/bitwarden/README.md) and associated images.

All commands in this package require the [Bitwarden CLI](https://bitwarden.com/help/article/cli/). The _Vault Status_, _Search Vault Items_, and _Copy First Password_ commands also require the [`jq` utility](https://stedolan.github.io/jq/). Install both via homebrew:

```sh
$ brew install bitwarden-cli jq
```

The only template command in the package is `log-in.template.sh`. If you use multifactor authentication to log in to Bitwarden, be sure to set the value of the `MFA_METHOD` variable. The [values available to use](https://bitwarden.com/help/article/cli/#enums) are:

| MFA Method        | Value |
|-------------------|:-----:|
| Authenticator App | 0     |
| Email             | 1     |
| Yubikey           | 3     |

If you _do not_ use multifactor authentication to access your Bitwarden vault, leave this variable set to `""`, and optionally remove the MFA Code argument (or leave it blank when logging in). Enabling multifactor authentication is encouraged.

## Checklist


commands/password-managers/bitwarden/copy-first-password.sh
```diff
@@ -0,0 +1,51 @@
...
+# Optional parameters:
+# @raycast.packageName Bitwarden
+# @raycast.icon images/bitwarden.png
+# @raycast.argument1 { "type": "text", "placeholder": "Query" }
+#
+# Documentation
+# @raycast.author Phil Salant
+# @raycast.authorURL https://github.com/PSalant726
+# @raycast.description Search all items in a Bitwarden vault, and copy the password of the first matching item to the clipboard.
+
```
unnamedd, line 23: As we have some dependencies like `bitwarden` and `jq`, will make more sense if we don't assume anything and make a check before the use. i.e:

```shell
if ! command -v bw &> /dev/null; then
  echo "bw command is required (https://bitwarden.com/help/article/cli/).";
  exit 1;
fi

if ! command -v jq &> /dev/null; then
  echo "jq command is required (https://stedolan.github.io/jq/).";
  exit 1;
fi
```

Add that piece of code in the scripts you created for Bitwarden will avoid the user to receive a strange error and you have more control.

commands/password-managers/bitwarden/copy-first-password.sh
```diff
@@ -0,0 +1,51 @@
...
+# Dependencies:
+#   1. The Bitwarden CLI: https://bitwarden.com/help/article/cli/
+#   2. The `jq` utility: https://stedolan.github.io/jq/
+#
+# Install via homebrew: `brew install bitwarden-cli jq`
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Copy First Password
```
unnamedd, line 11: `Copy First Password` brought me confusion.
Maybe if we add here a word `Matching` like you did in the description possibly will avoid confusions. What do you think?
```suggestion
# @raycast.title Copy First Matching Password
```

commands/password-managers/bitwarden/search-vault-items.sh
```diff
@@ -0,0 +1,47 @@
...
+  session_args="--session $token"
+fi
+
+bw unlock --check $session_args > /dev/null 2>&1
+unlocked_status=$?
+
+if [ $unlocked_status -ne 0 ]; then
+  echo "Vault is locked. Use the 'Log In' or 'Unlock' commands to enable searching."
+  exit 0
+fi
```
unnamedd, line 39: This piece of code you are using in two different places, what do you think to instead of copy and paste twice, create a separate shell file to receive some parameters and make that verification? It will create one more file, but if you decide to have a third place for that validation, you only need to make a call instead of replicate one more time the code.

unnamedd (changes requested):
Hey @PSalant726, the Script Command looks great to me. There are some comments but nothing really special! 

Thanks for the contribution, pal!

commands/password-managers/bitwarden/copy-first-password.sh
PSalant726: Oh wow I can't believe I forgot to include this - good catch!

commands/password-managers/bitwarden/search-vault-items.sh
PSalant726: In general I agree that there is tremendous value in keeping code DRY, but I think this `script-commands` repository is a special case. I intentionally repeated this code (and several other blocks of code) in multiple scripts in order to keep each script portable, and to keep the package as a whole from creating undue dependency on it itself. In other words, each script is fully functional on it's own, without any other scripts (or helper files) in the package. Also, I don't think the repeated code is so complex that it requires abstraction, so I would prefer to keep things this way.

unnamedd: Okay, I understood your point and make sense! Thanks for explain your point Phil!

commands/password-managers/bitwarden/search-vault-items.sh
```diff
@@ -0,0 +1,57 @@
...
+#
+# Documentation
+# @raycast.author Phil Salant
+# @raycast.authorURL https://github.com/PSalant726
+# @raycast.description Search all items in a Bitwarden vault.
+
+if ! command -v bw &> /dev/null; then
+  echo "The Bitwarden CLI is not installed (https://bitwarden.com/help/article/cli/)."
+  echo "Install via Homebrew with 'brew install bitwarden-cli'"
+  exit 1
```
unnamedd, line 28: In the previous file, your verification is a little different, would you mind to keep it consistence between the files? 

Previous file:
```shell
if ! command -v bw &> /dev/null; then
   echo "The Bitwarden CLI is not installed."
   exit 1
 fi
```

commands/password-managers/bitwarden/unlock-vault.sh
```diff
@@ -0,0 +1,49 @@
...
+#
+# Documentation
+# @raycast.author Phil Salant
+# @raycast.authorURL https://github.com/PSalant726
+# @raycast.description Unlock an authenticated Bitwarden vault.
+
+if ! command -v bw &> /dev/null; then
+  echo "The Bitwarden CLI is not installed."
+  exit 1
+fi
```
unnamedd, line 36: I made a comment above regarding the consistence of the messages. I do think that the one more descriptive is better to use than a simple one like this. What do you think, Phil?

commands/password-managers/bitwarden/vault-status.sh
```diff
@@ -0,0 +1,83 @@
...
+# @raycast.authorURL https://github.com/PSalant726
+# @raycast.description Display the authentication and lock status of the user's Bitwarden vault.
+
+if ! command -v bw &> /dev/null; then
+  echo "⚠️  The Bitwarden CLI is not installed"
+  exit 1
+elif ! command -v jq &> /dev/null; then
+  echo "⚠️  The jq utility is not installed"
+  exit 1
+fi
```
unnamedd, line 41: Another inconsistence.

unnamedd (changes requested):
Hey Phil, I made a second review and got some small things to be changed.

commands/password-managers/bitwarden/search-vault-items.sh
PSalant726: The verification output changes based on the `mode` of the script, in order to best take advantage of the way it would appear to the user. It's consistent between all `silent`, `inline`, and `fullOutput` commands.

unnamedd: Great! I didn't pay attention to the mode used and I agree with you that make more sense the way you've implemented. Looks great for me!

unnamedd:
No further comments in this PR, Phil! From my side, it is approved!

unnamedd (approved):

PSalant726:
Ok, I'm done tinkering with this. I won't push any more commits unless there are requested changes 😅

fe9lix (approved):
Pretty comprehensive collection of scripts and docs, thanks! I've just skimmed it and leave the main review up to @unnamedd.

PR #215, opened 2021-02-01
Add a script command to run an arbitrary shell command in iTerm

andreiborisov:
## Description

This PR adds a new script command which allows running of an arbitrary shell command in iTerm.

It executes the command in the existing window and tab unless there is already a running command there like `ssh` or `top` in which case it opens a new tab. It also has a variable that tweaks the script to open a new window instead of a tab.

## Type of change

- [x] New script command

## Video

https://www.icloud.com/iclouddrive/0ik_r5f285Va2HIHCYcGtCL5w#CleanShot_2021-02-01_at_17.30

iTerm

## Checklist


fe9lix (changes requested):
hey @andreiborisov, thanks! could you add a runtime check whether iTerm is installed on the user's machine? also see "Scripts that require installation of runtimes and dependencies" in https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md

andreiborisov:
> Could you add a runtime check whether iTerm is installed on the user's machine?

Are we sure we need a runtime check? I don’t see any checks in other script commands intended to control a particular app, e.g., https://github.com/raycast/script-commands/blob/master/commands/apps/amphetamine/amphetamine-on.applescript

fe9lix:
Fair point – in that case I think it's pretty clear already from the script command name that this only works with iTerm + it's in the "apps" folder, so should be fine.

fe9lix (approved):

PR #222, opened 2021-02-06
Update get external IP scripts to use ipify.org API instead of ifconfig.co

alexruf:
A CAPTCHA prevents the script from working with ifconfig.co API.
Therefore, the script now uses the ipify.org API.

### DIFF
--- a/commands/developer-utils/ip/get-external-ip-v4.sh
+++ b/commands/developer-utils/ip/get-external-ip-v4.sh
@@ -13,5 +13,5 @@
 # Documentation:
 # @raycast.description Copies the external IPv4 to the clipboard.
 
-ip=$(curl -4 -s -m 5 https://ifconfig.co)
+ip=$(curl -4 -s -m 5 https://api.ipify.org)
 echo $ip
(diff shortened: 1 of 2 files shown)


unnamedd:
Hi @alexruf, thanks for your first contribution updating the API to get the external IP, but I need to ask: in which case you received a captcha? I made a test here and everything is working to me. 

This is important to know to prevent others to have the same problem you had.

alexruf:
The captcha appears every time, no matter if I open the URL in the Browser or call the API via curl. It seems to come from Cloudflare.

unnamedd:
I'd argue the problem would be some kind of restriction to your country applied by CloudFlare, but as I'm in Germany, like you, this must be discarded. 

I don't have any problem with the replacement, we just need to ensure that no one will have that problem in the future.

unnamedd:
@thomaspaulmann @PitNikola @fe9lix any concerns from your side?

crstauf:
ifconfig.co appears to also be served by Cloudflare, so could happen there as well.

PitNikola:
No problem from our side as long as it works. Let's see if somebody complains 😅

PR #224, opened 2021-02-07
Color Conversion command

quelhasu:
## Description

Converts colors to a desired format.

## Type of change

- [x] New script command

## Screenshot

https://ibb.co/bBf0YJ4
https://ibb.co/n6nBngV

## Checklist


crstauf:
Note: partial duplicate of scripts added in #24.

quelhasu:
Not really, you can convert a color to (rgb, rgba, hex) and vice versa.

crstauf:
Hence the use of the word "partial".

commands/conversions/color-conversion.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Color Conversion
+# @raycast.mode fullOutput
+# @raycast.packageName Conversions
+
+# Optional parameters:
+# @raycast.icon 🎨
+# @raycast.currentDirectoryPath ~
```
unnamedd, line 11: As you are not acting in any specific directory, this meta tag isn't useful.

commands/conversions/color-conversion.sh
```diff
@@ -0,0 +1,64 @@
...
+# @raycast.needsConfirmation false
+# @raycast.argument1 { "type": "text", "placeholder": "Color (e.g. #FFEEFF)" }
+# @raycast.argument2 { "type": "text", "placeholder": "New format (e.g. RGB)" }
```
unnamedd, line 14: I was wondering here, how the user might know which options are available? Here I will do a suggestion. You don't have to accept this one, but will be nice if you check if that placeholder will have a nice width and then decide to apply or not. What do you think?

```suggestion
# @raycast.argument2 { "type": "text", "placeholder": "New format (e.g. RGB | RGBA | HEX)" }
```

commands/conversions/color-conversion.sh
```diff
@@ -0,0 +1,64 @@
...
+rgba)
+  transformed_color=$(printf "rgba(%d,%d,%d,1)" 0x${hex:0:2} 0x${hex:2:2} 0x${hex:4:2})
+  ;;
+
+hex)
+  transformed_color=$(echo "#${hex}")
+  ;;
+
+*)
+  echo "Select a correct format!"
```
unnamedd, line 57: Here, will be nice if you inform which formats are available. Maybe following my suggestion in the argument2.

unnamedd (changes requested):
Hi @quelhasu, despite the fact that part of your Script Command is duplicated, I liked it. 

The review is done, nothing really special. 
Thanks for your submission.

commands/conversions/color-conversion.sh
quelhasu: Yes it's a good idea, on the other hand we should do the same for the first argument because the user can input other formats (hex, rgb, rgba).

unnamedd: Indeed, I liked the idea! 
Somewhere in time, we should have in the argument a kind a list with pre-defined items and this kind of workaround won't be needed anymore, but until there, I guess this solution should solve the "problem".

commands/web-searches/search-gender-in-chosic.applescript
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/osascript
```
unnamedd, line 1: Don't forget to delete this file @quelhasu. It belongs to another PR.

unnamedd (approved):
Thanks for the changes @quelhasu, it’s approved. ✅ 
cc @thomaspaulmann  @PitNikola @fe9lix

PitNikola:
🎨 🎉

PR #229, opened 2021-02-09
URL shortener using bitly

gniting:
## Description
Script command that does URL shortening using Bitly.
* Will require the user to generate an access token on their bitly account. This way they can get metrics on the short URLs created. 
* No dependencies, uses curl to validate the input URL.

## Type of change
- [X] New script command

## Screenshot
image

None

## Checklist


commands/browsing/shorten-url-bitly.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
unnamedd, line 1: Due to the need of additional modification in this Script Command, please rename it to `shorten-url-bitly.template.sh`.

mattisssa:
Looks great! Thanks for contribution 👍

unnamedd (changes requested):
Hi @gniting, your Script Command looks pretty okay to me. You only need to make a small change (which is commented below) and then it will be approved! 

Thanks a lot for your contribution, one more time!

commands/browsing/shorten-url-bitly.sh
gniting: Done!

unnamedd (approved):
Thanks @gniting!

It's approved! 
@thomaspaulmann @PitNikola @fe9lix

PR #230, opened 2021-02-13
add script for baidu-translate, google-translate, jingdong

sunbufu:
## Description

1. search in Jingdong
2. translate with Baidu
3. translate with Google

## Type of change

- [x] New script command

## Checklist

- [x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

PR #231, opened 2021-02-13
Use wikipedia.org as default

asantos00:
## Description

Changes wikipedia search script to use wikipedia.org instead of chinese by default (zh)

## Type of change

## Checklist

### DIFF
--- a/commands/web-searches/search-in-wikipedia.sh
+++ b/commands/web-searches/search-in-wikipedia.sh
@@ -10,6 +10,6 @@
 
 # Optional parameters:
 # @raycast.icon images/wikipedia.png
-# @raycast.argument1 { "type": "text", "placeholder": "关键词...", "percentEncoded": true }
+# @raycast.argument1 { "type": "text", "placeholder": "Search term...", "percentEncoded": true }
 
-open "https://zh.wikipedia.org/w/index.php?search=$1"
+open "https://wikipedia.org/w/index.php?search=$1"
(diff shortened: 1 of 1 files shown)


unnamedd (approved):
Hi @asantos00, 
thanks for your update in your Wikipedia Script Command! 

It's approved!

PitNikola (changes requested):
@asantos00, I was wondering if it would make more sense to introduce a new script command for searching in english wikipedia instead of changing this one. I would assume that it was @Jax0rz's intention to use Chinese version of Wikipedia.

@asantos00, what do you think about renaming this script to "search-in-wikipedia-chinese.sh" and introducing a separate one?

unnamedd:
> @asantos00, I was wondering if it would make more sense to introduce a new script command for searching in english wikipedia instead of changing this one. I would assume that it was @Jax0rz's intention to use Chinese version of Wikipedia.
> 
> @asantos00, what do you think about renaming this script to "search-in-wikipedia-chinese.sh" and introducing a separate one?

@PitNikola I wrote a comment regarding languages in another PR that make sense to bring here, check it out and let me know what you think: https://github.com/raycast/script-commands/pull/230#discussion_r576296377

asantos00:
Would you be fine if I add a "domain" variable to the script (that can be chinese by default)?

I think this would fit @unnamedd suggestion.

The only issue I'm having is about the placeholder, as I think we can't use interpolation in comments. It was 关键词... before and I changed it to Search term...

What do you suggest doing here?

I'm pushing my variable with the domain variable, happy to add more stuff.

PitNikola:
Yeah, I think it's a good middle ground, thanks @asantos00. Let's ship it like this.

@unnamedd, good point about internationalization, we need to think about some guidelines for it. I think this idea with template might be a reasonable solution.

PR #239, opened 2021-02-18
New script: run arbitrary bash command and return output in Raycast

BoyangMichael:
## Description

Title:
run-bash-command.sh      

Description:
Run arbitrary bash command and show output in Raycast window.  

Usage:
`> <dir> <command> `  
              
Arguments:          
- `<dir>` the directory where your command will be executed. If blank, it uses current finder directory.       
- `<command>` Your arbitrary bash commands (oneliner). If blank, it opens the Terminal app and cd to `<dir>`.  

_**ATTENTION**_
I'm not sure which category it falls, so I put it under `apps/Raycast` because it return output in Raycast window. Please correct me if needed.

## Type of change

- [x] New script command

## Screenshot

run-bash-command_screenshot_1
run-bash-command_screenshot_2

None.

## Checklist


commands/apps/Raycast/run-bash-command.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
unnamedd, line 1: Please, move this script command to `developer-utils` folder.

commands/apps/Raycast/run-bash-command.sh
```diff
@@ -0,0 +1,55 @@
...
+# @raycast.authorURL https://github.com/BoyangMichael
+#
+# Optional parameters:
+# @raycast.icon images/run-bash-command.png
+# @raycast.currentDirectoryPath ~
+# @raycast.packageName Bash Command
+# @raycast.argument1 { "type": "text", "placeholder": "dir (blank: finder dir)", "optional": true }
+# @raycast.argument2 { "type": "text", "placeholder": "cmd (blank: open Terminal)", "optional": true }
+
+dir="${1/#\~/$HOME}"
```
unnamedd, line 34: Raycast execute Script Commands in a non-login environment and I wonder if that variable will be filled or not.

commands/apps/Raycast/run-bash-command.sh
```diff
@@ -0,0 +1,55 @@
...
+#                   * If blank, it uses current finder directory.             #
+#         <command> Your arbitrary bash commands (oneliner).                  #
+#                   * If blank, it opens the Terminal app and cd to <dir>.    #
+#                                                                             #
+###############################################################################
+#
+#
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title >
```
unnamedd, line 21:
```suggestion
# @raycast.title Run Command
```

commands/apps/Raycast/run-bash-command.sh
```diff
@@ -0,0 +1,55 @@
...
+# @raycast.description Run arbitrary bash command and return output in Raycast.
+# @raycast.mode fullOutput
+# @raycast.author Boyang Xu
+# @raycast.authorURL https://github.com/BoyangMichael
+#
+# Optional parameters:
+# @raycast.icon images/run-bash-command.png
+# @raycast.currentDirectoryPath ~
```
unnamedd, line 29: I guess your Script Command may have a conflict with the use of this metatag and the [first argument](https://github.com/raycast/script-commands/pull/239/files?short_path=b708721#diff-a4bf8857bf1c60a3f3a4965f0953409d325414e8dd7db37fde4db5e4ce59ad95R34).

unnamedd (changes requested):
Hi @boyangmichael, welcome aboard!

I made some comments in your PR, part of it brought me some questions which is nice to have them clear before the approval.

Thanks for your first contribution!

commands/apps/Raycast/run-bash-command.sh
BoyangMichael: Hi @unnamedd ! Thanks a lot for your revision! I will reply to your 4 comments together here.

1.  Moved to `developer-utils`.
2. Tested: `$HOME` is present even under non-login shell.
3. The reason why I use `>` instead `Run command` as the title is that Raycast alias does not accept special characters such as `>`. The use of verbose title without possibility of setting a simple alias would make typing difficult. My idea was to avoid hitting keyboard too much, so a one-character alias/title like `>` would be faster than others (e.g., `cmd`). If the verbose title is compulsory, I will change it. Just let me know.
4. Removed the metatag `# @raycast.currentDirectoryPath ~`.

Thanks again for your comments! Have a nice evening!
Boyang

unnamedd: Hi @BoyangMichael, thanks a lot for your quick act an answers. 

I understand your point regarding the title, the point is this one here in our [guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#metadata-convention): 
```txt
Title: Raycast's UI adopts title-cased strings for all command titles as per Apple's Human Interface Guidelines. Please make sure your command title follows this pattern to look good between other commands.
```

I wouldn't say it is compulsory, but I'd argue that it's more friendly for those who will see your Script Command in the list and will see only the sign `>`, it's not self-explainatory. Having it properly named, will make it more understandable even if the user doesn't know what's going to be done in that Script Command. It's a matter of a good naming for others. :D

Thiago

BoyangMichael: Hi @unnamedd , thanks for the detailed explanation! Now I understands. I have just changed the tile to `Run Command`.

Thanks for the help!
Boyang

unnamedd: You're more than welcome! Feel free to join us in our [Slack community](https://raycast.com/community), will be awesome to have you there! ;)

unnamedd (approved):
Looks good to me! It's approved! ✅ 
@thomaspaulmann @PitNikola

commands/developer-utils/run-bash-command.sh
```diff
@@ -0,0 +1,54 @@
...
+# @raycast.description Run arbitrary bash command and return output in Raycast.
+# @raycast.mode fullOutput
+# @raycast.author Boyang Xu
+# @raycast.authorURL https://github.com/BoyangMichael
+#
+# Optional parameters:
+# @raycast.icon images/run-bash-command.png
+# @raycast.packageName Bash Command
+# @raycast.argument1 { "type": "text", "placeholder": "dir (blank: finder dir)", "optional": true }
+# @raycast.argument2 { "type": "text", "placeholder": "cmd (blank: open Terminal)", "optional": true }
```
thomaspaulmann, line 31: @BoyangMichael how about switching those params? Then you could run simple commands that don't care about the `cwd` straight away without to tab into the next argument field.

Also, I would drop the hint in the brackets and make the command mandatory.
```suggestion
# @raycast.argument1 { "type": "text", "placeholder": "Command" }
# @raycast.argument2 { "type": "text", "placeholder": "Directory (Default: Finder)", "optional": true }
```

BoyangMichael: Hi @thomaspaulmann ! Good advice for switching the parameters! Thanks. I will switch it.

Concerning the change to make the command field mandatory, it will lead to loss of functions/actions. In the script, if the command field is blank, it will open terminal at the specified directory. If the directory is also blank, then it will open terminal at the directory of current Finder window. These functions will be lost due to this change.

BoyangMichael: Hi @thomaspaulmann ! I have just committed the changes: switched order of parameters but keep two arguments optional. Thanks for reviewing!

Have a nice evening,
Boyang

thomaspaulmann: Makes sense. Thanks for adding the changes. Gonna merge it now.

PR #254, opened 2021-02-23
feat: add undraw to web searches

AsterYujano:
## Description

Add undraw to the search folder: https://undraw.co/

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

none

none

## Checklist


commands/web-searches/undraw.sh
```diff
@@ -0,0 +1,12 @@
...
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Search in Undraw
+# @raycast.mode silent
+# @raycast.packageName Web Searches
+
+# Optional parameters:
+# @raycast.icon images/undraw.png
+
+open "https://undraw.co/search"
```
thomaspaulmann, line 12: Thanks for the contribution @AsterYujano. Is there a way to specify a query parameter that you could pass to Undraw. Then the search could be more efficient and you would see result straigth away. This is how other web searches work.

thomaspaulmann (changes requested):
Putting back in your queue to check if you can pass a search query to the web search.

commands/web-searches/undraw.sh
AsterYujano: Hello,

Sadly after a quick search, I don't think undraw allows a normal GET parameter to run a search.
The search is done by POST request with a payload "{query: "mysearch"}".

I don't think it is possible to pass parameters like the other searches, as it is not a GET.

What do you think?

unnamedd: You’re right @AsterYujano. Before write any comments here, I was also checking how unDraw works and as they are not rendering the result based on the URL parameters, I don’t think will be possible to inject the value to trigger the search. So sad! :(

thomaspaulmann:
Since this web search doesn't take search query, it is simply a bookmark. We usually don't add simple bookmarks to this repository because there are too many and it would explode this repo. I hope that's understandable. @AsterYujano if you find a way to pass a query to the search, please feel free to re-open the PR.

PR #264, opened 2021-02-27
Improve speedtest.sh

Pondorasti:
## Description

When using the [`speedtest.sh`](https://github.com/raycast/script-commands/blob/master/commands/dashboard/speedtest.sh) command you need to install the Speedtest CLI. However, you also need to accept their license in order to use it. Therefore, I have added a new installation step to the command to make it more clear.

Additionally, the `speedtest-logo.png` was `1024px`, which made it look pixelated because of all the downscaling. Therefore, I resized the image to 32px.

## Type of change

## Checklist


thomaspaulmann:
That's wonderful! Thanks for your first contribution @Pondorasti, off to many more 🎉

PR #275, opened 2021-03-04
script to show and manage surfshark VPN connection

jordicl:
## Description

These 3 scripts are used to manage your Surfshark VPN connection. It can show the current status inline, connect and disconnect the VPN.

## Type of change

- [x] New script command

## Screenshot

Screenshot 2021-03-04 at 23 00 31

## Dependencies / Requirements

This script depends on the `vpnutil` command line utility. This utility can be downloaded here: https://github.com/Timac/VPNStatus/releases/tag/1.0. Please make sure it's added to your `$PATH`.

## Checklist

- [x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

PR #277, opened 2021-03-05
New Scripts: VPN

Pondorasti:
## Description

3 new template scripts for managing your VPN connection.
* `vpn-status.template.sh` - Dashboard script for quickly checking the status of your VPN
* `vpn-start.template.sh` - Starts VPN and has some internal logic to detect the connection was successful or not.
* `vpn-start.template.sh` - Stops VPN

## Type of change

- [x] New script command

## Screenshot

https://user-images.githubusercontent.com/32957606/110061069-c2292f00-7d1b-11eb-826c-613ffab7293b.mov

## Checklist


Pondorasti:
I just realized that there's a similar [PR](https://github.com/raycast/script-commands/pull/275), but more targeted towards SurfShark VPN. I guess great minds do think a like 🙃.

Also, I'm not so sure if my naming and documentation is inline with other template scripts. Let me know if any changes are needed.

jordicl:
> I just realized that there's a similar [PR](https://github.com/raycast/script-commands/pull/275), but more targeted towards SurfShark VPN. I guess great minds do think a like 🙃.

I like your solution better, as it doesn't has an external dependency. But unfortunately I did not get scutil to work. It always results in "No service". Googling for a bit I don't seem to be the only one. Perhaps related to the fact I'm using a 3rd party VPN tool?

Pondorasti:
I managed to get scutil working only if I would pass in all my credentials. Afterwards, I found out that I can use networksetup, which I think it grabs my VPN   config directly from keychain. As of now, I'm using scutit only to check the status of the connection.

Pondorasti:
@jordicl have you tried using my start/stop scripts with SurfShark?

jordicl:
> @jordicl have you tried using my start/stop scripts with SurfShark?

Yes, but unfortunately it doesn't work as it uses `scuti` and it does not pick up on my VPN configuration created by Surfshark. Same applies for NordVPN and ExpressVPN. I think they all use the same mechanism.

unnamedd:
Nice both of you to achieve almost the same result but using different tools!

unnamedd (changes requested):
Hi @Pondorasti, thanks for your VPN Script Commands! I'm sure a lot of people will use them!

The Scripts Commands themselves looks really good, however, I do think we should concentrate them all at the same place. As you are not using any external dependency as Jordi, let's move all the Script Commands to a folder `VPN` and place it inside of `System` folder.

Plus, I'd also like to add that would be great if you concentrate the configuration of the VPN in another file, i.e: `vpn-config.sh` and export from there, the VPN name and in each file, you run the config file before use the VPN variable. Let me show you what I had in mind:

file: `vpn-config.sh`
```shell
#!/usr/bin/env bash

export VPN_NAME="VPN (IKEv2)"

if [ -z "$VPN_NAME" ]; then
  echo "\$VPN_NAME is empty";
  exit 1;
fi
```

inside of (for example) `vpn-status.sh`:
```shell
source vpn-config.sh

VPN=$VPN_NAME
```

in that sense, we won't need to make all the Script Command files `.template`, only the config file.

If you need any help, will be a pleasure to help!

Pondorasti:
@unnamedd I just made a commit with the requested changes.

One thing though, I do think that the title for `vpn-status.sh` should be `VPN Status`. Might just be personal preference, but I feel it's a bit more clear.

**Current Behaviour**

Screenshot 2021-03-05 at 16 01 25

unnamedd:
The mode is `inline`... I agree with you, @Pondorasti. Let's revert that change then.

commands/system/vpn/vpn-start.sh
```diff
@@ -0,0 +1,46 @@
...
+    [ $loops -le $max_loops ]
+}
+
+networksetup -connectpppoeservice "$VPN"
+
+if poll_until_connected "$VPN"; then
+    echo "Connected to $VPN!"
+else
+    echo "Couldn't connect to $VPN"
+    scutil --nc stop "$VPN"
```
Pondorasti, line 45: I was also thinking about error handling, and I was wondering if I should return `exit 1` when the connection to a VPN fails.

I saw that Raycast has [error handling](https://github.com/raycast/script-commands#error-handling), but it is not clear to me if it should be used only if the script execution fails or the action itself is not successful. The latter would make more sense.

Pondorasti:
> The mode is `inline`... I agree with you, @Pondorasti. Let's revert that change then.

I just reverted the `title` by doing an amend commit.

commands/system/vpn/vpn-config.template.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# This is a template which needs further setup. Duplicate the file,
+# remove `.template.` from the filename, and set the name of
+# your VPN config from System Preferences.
+
+# Name of your VPN Config
+export VPN_NAME="VPN (IKEv2)"
```
unnamedd, line 8: Is this a default name? If not, let's remove it to fall in the condition below and show the error to the user.

commands/system/vpn/vpn-config.template.sh
```diff
@@ -0,0 +1,13 @@
...
+
+if [ -z "$VPN_NAME" ]; then
+  echo "\$VPN_NAME is empty";
```
unnamedd, line 11: Will make sense here to indicate in which file the user should configure the VPN name.

```suggestion
  echo "\$VPN_NAME is empty";
  echo "Please, rename it in \"vpn-config.sh\".";
```

commands/system/vpn/vpn-config.template.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
unnamedd, line 1: Also, the `.template` pattern will only work for the script commands itself, not here, then I suggest remove it from file name.

commands/system/vpn/vpn-start.sh
unnamedd: You can use the `exit 1;` here to report the error as well, no problem.

unnamedd (changes requested):

commands/system/vpn/vpn-config.template.sh
Pondorasti: That is the default name if you create VPN with an `IKEv2` config. There's 2 other types `Cisco IPSec` and `L2TP`. I think you are right and it's best to leave it empty.

commands/system/vpn/vpn-config.template.sh
Pondorasti: Makes sense. In this case, I will remove the template comments:
```
# This is a template which needs further setup. Duplicate the file,
# remove `.template.` from the filename, and set the name of
# your VPN config from System Preferences.
```

And update line 7 with this:
```
# Name of your VPN Config from System Preferences
```

commands/system/vpn/vpn-start.sh
```diff
@@ -0,0 +1,46 @@
...
+# Optional parameters:
+# @raycast.icon 📡
+
+# @Documentation:
+# @raycast.packageName VPN
+# @raycast.description Start VPN connection.
+# @raycast.author Alexandru Turcanu
+# @raycast.authorURL https://github.com/Pondorasti
+
+
```
Pondorasti, line 17: @unnamedd  how do you plan to make the `vpn-config.sh` discoverable, hence it will not show up in the big `README.md`? I could write a small text at the top of each file, similar to this one:

```suggestion
Requirements: download and configure `vpn-config.sh`
Afterward, place it in the same directory
```

unnamedd: There’s no need to make it discoverable in the README or even by Raycast, it will be used only by the files you wrote, the most important information to be presented in the documentation are already in the .sh files. Also, it will be downloaded with your file as well, because today, when you download the Script Commands today, you will automatically download the config file. Don’t worry about the config file. It is there just to reduce duplication of code and place where the user will configure the VPN name 😃. 

In the future, when a kind of Script Command Manager we might have a list of files related to that group of files (like here).

commands/system/vpn/vpn-status.sh
```diff
@@ -0,0 +1,24 @@
...
+# @raycast.author Alexandru Turcanu
+# @raycast.authorURL https://github.com/Pondorasti
+
+
+source vpn-config.sh
+VPN=$VPN_NAME
+
+status=$(scutil --nc status "$VPN" | sed -n 1p)
+
+echo "$status"
```
PSalant726, line 24: It would be nice to include the `$VPN` value in this output, so the dashboard command shows something like "Connected to `$VPN`".
```suggestion
if [ "$status" == "Connected" ]; then
  echo "$status to $VPN"
  exit 0
fi

echo "$status"
```

PSalant726 (approved):
Thanks for these! Just a minor usability suggestion, but not a blocker. LGTM!

commands/system/vpn/vpn-status.sh
Pondorasti: Interesting idea. From my usage, I'm always going to be connecting to the same VPN, which makes the echo message too verbose and repetitive over time. 

At the end of the day, it comes down to style and aesthetics. Works either way for me.

grzegorzkrukowski:
@unnamedd @Pondorasti - is there anything else you want to change in this PR before I merge it ?

I checked it and it looks good enough to be merged - we can do improvements in next PRs maybe ?

unnamedd:
Looks good to me as well @grzegorzkrukowski. We just need to wait to for the last change requested by @PSalant726 agreed by @Pondorasti. 

After the change, push the red button! 🤣

Pondorasti:
@grzegorzkrukowski kill it with 🔥!

PR #278, opened 2021-03-05
Update open-terminal-from-finder.applescript

rolandleth:
## Description

Added logic to check for window, selection and type of selection (alias, file, folder).

I would argue against `clear;` in the final command, but the original implementation not being mine, I've left it as is. My argument against it is that you might have output that you need, the most basic example being comparing two `ls` outputs.

## Type of change

- [x] Bug fix

## Checklist

### DIFF
--- a/commands/navigation/open-terminal-from-finder.applescript
+++ b/commands/navigation/open-terminal-from-finder.applescript
@@ -15,15 +15,41 @@
 # @raycast.authorURL https://github.com/japanese-goblinn
 
 tell application "Finder"
-    set pathList to (quoted form of POSIX path of (folder of the front window as alias))
-    set command to "clear; cd " & pathList
+    -- Check if there's a selection; works if there's a window open or not.
+    if selection is not {} then
+        set i to item 1 of (get selection)
+
+        -- If it's an alias, set the item to the original item.
+        if class of i is alias file then
+            set i to original item of i
+        end if
+
+        -- If it's a folder, use its path.
+        if class of i is folder then
+            set path to i
+        else
+            -- If it's an item, use its container's path.
+            set path to container of i
+        end if
+    else if (exists window 1) and current view of window 1 is in {list view, flow view} then
+        -- If a window exist, use its folder property as the path.
+        set path to folder of window 1
+    else
+        -- Fallback to the Desktop, as nothing is open or selected.
+        set path to path to desktop folder
+    end if
+
+    set command to "clear; cd " & quoted form of POSIX path of (path as alias)
 end tell
-  
+
 tell application "Terminal"
     if not (exists window 1) then reopen
-        activate
+
+    activate
+
     if busy of window 1 then
         tell application "System Events" to keystroke "t" using command down
     end if
+
     do script command in window 1
 end tell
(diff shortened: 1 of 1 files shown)


unnamedd (approved):
Looks good to me! 👍

grzegorzkrukowski:
Hey @rolandleth that's a quite advanced AppleScript ! Thanks for your contribution.

I am afraid we have problems running the script - we tried both Raycast and terminal directly.

The error we are getting is:
`Syntax Error`: `A end of line can’t go after this identifier.` for `set path to i` line.

But there are more syntax error inside the current version of the script - removing this line is triggering same errors on next lines.

rolandleth:
Hey! It was because the variable name `path` was conflicting with the keyword `path` (same reason I had to name the other variable `i` instead of `item` 🤦🏻‍♂️). Should work now, sorry for that 👍🏻

grzegorzkrukowski:
Thanks @rolandleth for updating the code - now it works. I will look more into it tomorrow, but I found some inconsistent behaviour of this script.
If I don't have terminal opened it launches terminal and opens an extra tab for it - otherwise it is using active tab.
Do you think that's something that will be easy to fix - to avoid opening two tabs when it needs to start terminal ? 

Also I agree about `clear` option. If we keep behaviour of always opening on active tab, we shouldn't really clear the current content - let's try to apply this change - if @japanese-goblinn doesn't have anything against it.

japanese-goblinn:
> Thanks @rolandleth for updating the code - now it works. I will look more into it tomorrow, but I found some inconsistent behaviour of this script.
> If I don't have terminal opened it launches terminal and opens an extra tab for it - otherwise it is using active tab.
> Do you think that's something that will be easy to fix - to avoid opening two tabs when it needs to start terminal ?
> 
> Also I agree about `clear` option. If we keep behaviour of always opening on active tab, we shouldn't really clear the current content - let's try to apply this change - if @japanese-goblinn doesn't have anything against it.
> 
> Another minor thing is that output message of this script is a bit useless - maybe we could print instead a path that was opened - or just simple confirmation message, instead of `tab 1 of window id 1231231`.
> Something as simple as: `do shell script "echo opened " & quoted form of POSIX path of (p as alias)` at the end can make it more friendly (maybe keeping right part after & as variable as it's used above)

I think that this is really nice improvement!

rolandleth:
Hmm, no idea why it opens a new tab for you, it never does for me, unless the current tab is busy. This part I didn't change from the original script either. The only idea I have is that it _might_ take too long for the initial tab to activate and it considers it busy. Could you try adding `delay 0.5` (or 2, just for testing purposes) after `activate`?

Agree with the message, question is if the script shouldn't be silent 🤔 (compact is too small for what it needs to print). The question about this, though, is won't it be redundant, now that we remove the `clear`? You'll have the `cd path` command that has be run right in front of you.

grzegorzkrukowski (changes requested):
Thanks for reasonable arguments @rolandleth.
As for separate tab - let's leave it as it is then.
As for output - let's remove `clear` and make the script really `silent` by adding `return ""` at the very end so there will be no popup - we will leave everything visible in terminal only.

grzegorzkrukowski:
Thanks @rolandleth for you effort ! Looks great.

PR #283, opened 2021-03-08
New Script: network-status.sh

Pondorasti:
## Description

System script that detects your current network connections. It can have the following states:
* `Ethernet`
* `Ethernet | <wifi-name>`
* `<wifi-name>`
* `No connection`

I'm open to any feedback 🙂. For the icon I have also considered these 2 emojis 🌐 📡, and the `wifi` segment was initially using this style `Wi-fi (<wifi-name>)`

## Type of change

- [x] New script command

## Screenshot

https://user-images.githubusercontent.com/32957606/110263534-debba600-7f6b-11eb-9d99-c0f6c8e535cb.mov

## Checklist

### DIFF
--- a/commands/system/network-status.sh
+++ b/commands/system/network-status.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Network Status
+# @raycast.mode inline
+# @raycast.refreshTime 1m
+
+# Optional parameters:
+# @raycast.icon 📶
+
+# @Documentation:
+# @raycast.packageName System
+# @raycast.description Get current network connections.
+# @raycast.author Alexandru Turcanu
+# @raycast.authorURL https://github.com/Pondorasti
+
+function get_wifi_ssid () {
+  service_info=$(/System/Library/PrivateFrameworks/Apple80211.framework/Versions/A/Resources/airport --getinfo)
+  ssid_line=$(echo "$service_info" | grep -v "BSSID" | grep "SSID")
+  ssid=$(echo "$ssid_line" | awk -F "(: )" '{print $2}')
+
+  echo "$ssid"
+}
+
+current_device=$(route get default 2>/dev/null | grep "interface" | awk '{print $2}')
+
+# Exit if there's no connection
+if [ -z "$current_device" ]; then
+  echo "No connection"
+  exit 0;
+fi
+
+service_info=$(networksetup -listnetworkserviceorder | grep "$current_device")
+service_name=$(echo "$service_info" | awk -F  "(, )|(: )|[)]" '{print $2}')
+wifi_ssid=$(get_wifi_ssid)
+network_status=""
+
+if grep -q "USB" <<< "$service_name"; then
+  network_status+="Ethernet"
+
+  # Check if also connected to Wi-fi
+  if [ -n "$wifi_ssid" ]; then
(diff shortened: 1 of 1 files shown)


grzegorzkrukowski:
🌐 could be too close to Network Application in System preferences, so Antenna Bars is a good pick.

Current version of the script is just checking connection to closest router.
Do you think extending the script to check if the internet connection is really there could be a good idea ? 
Maybe doing `ping` to google.
That will allow to display more detailed information: `Connected to $wifi_name without Internet Connection`.
I think most often user will check his status connection when things are not working - it would be great to give him a hint that router works, but next hop is failing.

Pondorasti:
@grzegorzkrukowski Technically I could do `ping -C 2 google.com`, which would try to ping google.com 2 times. In terms of my testing, I noticed that if my mac is connected both to Ethernet and Wi-fi, ethernet will always take priority (even if it has no internet connection).

I also liked your formatting. Now my question is, how would you handle the other states with Ethernet? Maybe `Connected to Ethernet and $wifi-name without Internet Connection`.

Alternatively, I could also do this where I have more granular control over the error:
* `Ethernet (No Internet Connection)`
* `Ethernet (No Internet Connection) | $wifi-name`
* `$wifi-name (No Internet Connection)`
* `No connection`

grzegorzkrukowski:
@Pondorasti pinging google sounds like a good plan.

As for formatting I think having what you proposed looks good - maybe we can even make it shorter and just show `No Internet` for that case.

grzegorzkrukowski:
Hey @Pondorasti are you planning to work on this PR in near days or should we merge it as it is for now and we can update this script in next iteration ? It's good enough to be merged as it is now, but if you plan to get it fully featured I will keep this PR for a bit longer.

Pondorasti:
@grzegorzkrukowski There's no need to delay it, I just pushed my changes for checking internet status 🙃. You can try it out yourself and let me know what you think.

I'm using this format, as we've discussed above:

> * `Ethernet (No Internet Connection)`
> * `Ethernet (No Internet Connection) | $wifi-name`
> * `$wifi-name (No Internet Connection)`
> * `No connection`

Screenshot 2021-03-11 at 10 29 05

grzegorzkrukowski:
Looking good. In terms of parsing `ping` results and doing `-c 2` - right now it seems that even if internet will be a bit crappy and 1 ping will fail and the other one will pass through - we will still show `No Internet`.
We can either do just 1 ping maybe - or we should maybe rather look for `ping statistics` and `"100.0% packet loss"` ?

Pondorasti:
Using this regexp `100.0% packet loss`  looks like the best way to handle internet connection.

Pondorasti:
I will have to take a look and see what's going on with the extra message from `ping`.

Pondorasti:
After a bit of digging around, I've come to the conclusion to redirect all output to `/dev/null` and just check for the returned code. 

I think this is how they work:
* `68` - cannot resolve hostname
* `2` - unreachable 
* `0` - at least one ping was successful 

With this implementation, if you loose internet connection in the middle of the ping test, where the first ping was successful, the script will tell you that you are connected. However, I think this is a highly unlikely scenario.

grzegorzkrukowski:
Wow really quick and deep research ! 

I think if we get at least one working ping out of 2 we send we can assume Internet is there.
Worst case it will update to `No Internet` 1 minute later ;)

PR #285, opened 2021-03-08
bitwarden: fix clipboard copy error that may include newline and message when copying password for a non-existing account

qeude:
## Description

This is fixing a bug that was adding a newline at the end of the copied password and a bug on displayed message when copying password for a non-existing account (example bellow)

## Type of change

- [ ] Bug fix

## Screenshot
The first call is performed before changes, the second one is before after my changes.
Screenshot 2021-03-08 at 19 01 54

## Checklist

- [ x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

### DIFF
--- a/commands/password-managers/bitwarden/copy-first-matching-password.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-password.sh
@@ -54,7 +54,7 @@ if [[ -z $name || -z $password || ( $name == "null" && $password == "null") ]]; 
   exit 1
 fi
 
-echo $password | tr -d " \t\n\r" | pbcopy
 unset $password
+echo -n $password | pbcopy
 echo "Copied the password for '$name' to the clipboard."
 exit 0
--- a/commands/password-managers/bitwarden/copy-first-matching-password.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-password.sh
@@ -50,8 +50,8 @@ if [ $unlocked_status -ne 0 ]; then
 fi
 
 item=$(bw list items --search $1 $session 2> /dev/null | jq ".[0] | { name: .name, password: .login.password }")
-name=$(echo $item | jq -e ".name") || notFoundError
-password=$(echo $item | jq -re ".password") || notFoundError
+name=$(echo $item | jq --exit-status ".name") || notFoundError
+password=$(echo $item | jq --raw-output --exit-status ".password") || notFoundError
 
 echo -n $password | pbcopy
 unset password
--- a/commands/password-managers/bitwarden/copy-first-matching-password.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-password.sh
@@ -54,7 +54,7 @@ if [[ -z $name || -z $password ]]; then
   exit 1
 fi
 
-echo $password | pbcopy
+echo $password | tr -d " \t\n\r" | pbcopy
 unset $password
 echo "Copied the password for '$name' to the clipboard."
 exit 0
--- a/commands/password-managers/bitwarden/copy-first-matching-password.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-password.sh
@@ -21,7 +21,7 @@
 # @raycast.authorURL https://github.com/PSalant726
 # @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
 notFound() {
-  echo -n "The query '$1' did not return a password."
+  echo "The query '$1' did not return a password."
   exit 1
 }
 
--- a/commands/password-managers/bitwarden/copy-first-matching-password.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-password.sh
@@ -20,6 +20,7 @@
 # @raycast.author Phil Salant
 # @raycast.authorURL https://github.com/PSalant726
 # @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
+
 notFound() {
   echo "The query '$1' did not return a password."
   exit 1
@@ -56,5 +57,4 @@ password=$(echo $item | jq --raw-output --exit-status ".password") || notFoundEr
 echo -n $password | pbcopy
 unset password
 echo "Copied the password for '$name' to the clipboard."
-unset name
 exit 0
--- a/commands/password-managers/bitwarden/copy-first-matching-password.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-password.sh
@@ -49,7 +49,7 @@ item=$(bw list items --search $1 $session 2> /dev/null | jq ".[0] | { name: .nam
 name=$(echo $item | jq ".name")
 password=$(echo $item | jq --raw-output ".password")
 
-if [[ -z $name || -z $password ]]; then
+if [[ -z $name || -z $password || ( $name == "null" && $password == "null") ]]; then
   echo "The query '$1' did not return a password."
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -49,12 +49,12 @@ item=$(bw list items --search $1 $session 2> /dev/null | jq ".[0] | { name: .nam
...
 password=$(echo $item | jq --raw-output ".password")
 
-if [[ -z $name || -z $password ]]; then
+if [[ -z $name || -z $password || ( $name == "null" && $password == "null") ]]; then
   echo "The query '$1' did not return a password."
   exit 1
 fi
 
-echo $password | pbcopy
+echo $password | tr -d " \t\n\r" | pbcopy
```
PSalant726, line 57: Per the [`jq` documentation](https://stedolan.github.io/jq/manual/#Invokingjq), we should probably be using `--join-output` instead of `--raw-output` on L50, instead of trying to script the `\n` away like this. I didn't realize `--join-output` was an option when I originally wrote this script.

commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -49,12 +49,12 @@ item=$(bw list items --search $1 $session 2> /dev/null | jq ".[0] | { name: .nam
 name=$(echo $item | jq ".name")
 password=$(echo $item | jq --raw-output ".password")
 
-if [[ -z $name || -z $password ]]; then
+if [[ -z $name || -z $password || ( $name == "null" && $password == "null") ]]; then
```
PSalant726, line 52: We should be able to use the `jq --exit-status` option and check the value on/after L49, instead of adding these additional conditions.

PSalant726 (changes requested):
Thanks for these fixes! Just some suggestions for a cleaner implementation.

commands/password-managers/bitwarden/copy-first-matching-password.sh
qeude: I did investigate a bit, and I finally found that the added newline isn't caused by the `jq` command but by the `echo` one.
I think the following code would be the cleanest way to resolve this to avoid the added newline at the end from the `echo` command as documented [here](https://linuxcommand.org/lc3_man_pages/echoh.html).
```suggestion
echo -n $password | pbcopy
```

commands/password-managers/bitwarden/copy-first-matching-password.sh
qeude: I fixed this by using `jq --exit-status` in [this commit](https://github.com/raycast/script-commands/pull/285/commits/91fa07beae98f396f604ec5e0f6fb50e75006e21). It make all the checking stuff cleaner.

commands/password-managers/bitwarden/copy-first-matching-password.sh
PSalant726: Good looks! Yes, let's go with `echo -n`.

commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -46,15 +50,11 @@ if [ $unlocked_status -ne 0 ]; then
...
 item=$(bw list items --search $1 $session 2> /dev/null | jq ".[0] | { name: .name, password: .login.password }")
-name=$(echo $item | jq ".name")
-password=$(echo $item | jq --raw-output ".password")
-
-if [[ -z $name || -z $password ]]; then
-  echo "The query '$1' did not return a password."
-  exit 1
-fi
+name=$(echo $item | jq -e ".name") || notFoundError
+password=$(echo $item | jq -re ".password") || notFoundError
```
PSalant726, line 54: When available, it's always a best practice to use the long-form of command options when coding them into a script. This preserves clarity, and in this case there aren't enough options to require the commands to overflow onto additional lines.
```suggestion
name=$(echo $item | jq --exit-status ".name") || notFoundError
password=$(echo $item | jq --raw-output --exit-status ".password") || notFoundError
```

commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -46,15 +50,11 @@ if [ $unlocked_status -ne 0 ]; then
...
 
-echo $password | pbcopy
-unset $password
+echo -n $password | pbcopy
+unset password
```
PSalant726, line 57: Does this need to remain `$password`?

commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -46,15 +50,11 @@ if [ $unlocked_status -ne 0 ]; then
...
 echo "Copied the password for '$name' to the clipboard."
-exit 0
+unset name
+exit 0
```
PSalant726, line 60: Please preserve the newline at the end of the file, to avoid superfluous diffs if/when others edit it.

commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -46,15 +50,11 @@ if [ $unlocked_status -ne 0 ]; then
...
+name=$(echo $item | jq -e ".name") || notFoundError
+password=$(echo $item | jq -re ".password") || notFoundError
 
-echo $password | pbcopy
-unset $password
+echo -n $password | pbcopy
+unset password
 echo "Copied the password for '$name' to the clipboard."
-exit 0
+unset name
```
PSalant726, line 59: This shouldn't contain sensitive info, and the next line `exit`s the script, so it doesn't need to be `unset`.

commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -20,6 +20,10 @@
 # @raycast.author Phil Salant
 # @raycast.authorURL https://github.com/PSalant726
 # @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
+notFound() {
+  echo -n "The query '$1' did not return a password."
```
PSalant726, line 24: Not a big deal, but since the next line will always exit the script, `-n` is unnecessary and just adds extra complexity.
```suggestion
  echo "The query '$1' did not return a password."
```

PSalant726 (changes requested):
Nice changes! Just a few quality-of-life suggestions and this should be good to merge!

commands/password-managers/bitwarden/copy-first-matching-password.sh
qeude: It seems that with the `$` it doesn't unset properly the variable, which might be a non wanted behaviour.

commands/password-managers/bitwarden/copy-first-matching-password.sh
qeude: It seems more consistent to me to explicitly unset the var, even if it's not sensitive. But I'm ok to revert this change.

commands/password-managers/bitwarden/copy-first-matching-password.sh
PSalant726: Yes, my mistake. Good call.

commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -20,6 +20,10 @@
 # @raycast.author Phil Salant
 # @raycast.authorURL https://github.com/PSalant726
 # @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
+notFound() {
```
PSalant726, line 23: Just a nitpick, but I'd like it if a blank line separated the documentation from the code here.
```suggestion

notFound() {
```

commands/password-managers/bitwarden/copy-first-matching-password.sh
PSalant726: It's not a necessary line, so let's remove it for simplicity.

PSalant726 (changes requested):
Should be good after these changes. Thanks for sticking with it!

qeude:
Should be ok now @PSalant726 :)

PSalant726 (approved):
Looks good, thanks!

PR #287, opened 2021-03-09
Added new script and icon for simple countdown

marcocebrian:
## Description

Simple AppleScript script to receive a notification with sound after x minutes with an optional title.

## Type of change

- [X ] New script command

## Screenshot

Screenshot

No dependencies.

## Checklist


commands/system/countdown.applescript
```diff
@@ -0,0 +1,31 @@
...
+# Optional parameters:
+# @raycast.icon images/clock.png
+# @raycast.argument1 { "type": "text", "placeholder": "minutes" }
+# @raycast.argument2 { "type": "text", "placeholder": "Title", "optional": true }
+
+
+# @Documentation:
+# @raycast.description Simple minutes countdown
+# @raycast.author Marco Cebri�n
+
```
Pondorasti, line 18: I would suggest to also add your GitHub account to the `authorURL` field.
```suggestion
# @raycast.authorURL https://github.com/marcocebrian

```

marcocebrian: Done! Thanks

grzegorzkrukowski (changes requested):
Thanks for your first contribution.

First of all I think that script should be named differently: maybe "Schedule Notification".
Countdown suggests that it will be showing remaining time allowing user to see how much time is left till countdown finishes.

To be more consistent also with naming in Raycast I suggest to uppercase parameters: `In Minutes`.

Would be also great to get some kind of confirmation at the end of the script - making it `silent` still allows you to show a Toast message for a user with confirmation, for example: "Scheduled notification $TITLE at 16:00AM" or something like that.
Right now as there is also no way to remove and managed those notifications user can get a feeling that something went wrong with scheduling.

marcocebrian:
Changes executed:

- [x] Renamed script and adjusted English grammar.
- [x] Renamed parameters for consistency with Raycast.
- [x] Implemented toast message. (Small problem here)


The problem with the toast is that:
In AppleScript, when you `log` text for the toast to capture the message, it is done asynchronously and the log is not executed until the delay is executed (end of waiting time). So, the toast comes at the end.

I have tried also with `do shell script` with identical results. It is executed after the delay, even in the line of code is before.

Any help to force it to "flush" or execute the same thing synchronized?
Thanks

marcocebrian (commented):
Changes executed:

- [x] Renamed script and adjusted English grammar.
- [x] Renamed parameters for consistency with Raycast.
- [x] Implemented toast message. (Small problem here)


The problem with the toast is that:
In AppleScript, when you `log` text for the toast to capture the message, it is done asynchronously and the log is not executed until the delay is executed (end of waiting time). So, the toast comes at the end.

I have tried also with `do shell script` with identical results. It is executed after the delay, even in the line of code is before.

Any help to force it to "flush" or execute the same thing synchronized?
Thanks

grzegorzkrukowski:
Hey @marcocebrian - I am afraid that this PR with current implementation cannot be merged.

We don’t support long-running scripts in Raycast and this one can run even for few hours if you setup long enough notification delay. This is why it's also hard for you to get properly Toast message being displayed.
Also you cannot really rely on trusting that it will be displayed once you close Raycast or simply script will get killed by running for too long.

They only solution I see to have notifications would be to integrate it with an API that allows you to schedule notification and return straight away - but I am not sure if there's any - I didn't find it for AppleScript at least.

Let me know if you find a solution that will not use `delay` or `sleep` approach - if not, I am afraid we have to close this one.

marcocebrian:
I understand you point @grzegorzkrukowski, however, it find this functionality very useful. Any possibility of you integrating it in your main app? To Schedule a notification from a swift Mac OS X app is easy, efficient and safe, even with the App closed or killed.

I will explore the possibility of doing it from a Swift script... but I am afraid it will require a full app to be enqueued into the Notification Center...

grzegorzkrukowski:
We don’t have immediate plans to add notifications to Raycast. At some point, we might unblock long-running tasks such as your script but we can’t give an estimate on it right now.

PR #288, opened 2021-03-09
Quit app script

rolandleth:
## Description

Script that takes in a name or a process id and terminates it. I, personally, use it with "kill process" instead of "quit app", but not everyone is technical, so I went with a more friendly name for the public version. But I'm up for debate on this one, because I'm not totally sure how it should go for the public. I also tried "Terminate app", but that also didn't feel right.

## Type of change

- [x] New script command

## Screenshot

Bildschirmfoto 2021-03-09 um 8 16 45 AM

## Checklist

### DIFF
--- a/commands/system/quit-app.swift
+++ b/commands/system/quit-app.swift
@@ -25,7 +25,7 @@ if let processId = Int(argument) {
 			.runningApplications
 			.first(where: { $0.processIdentifier == processId })
 	else {
-		print("Couldn't find app with process id \(processId)")
+		print("No apps with process id \(processId)")
 		exit(1)
 	}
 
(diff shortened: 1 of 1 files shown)


grzegorzkrukowski (changes requested):
Let's try to make "killing" less destructive - as I described in my comment, killing "Finder" caused some really bad side effects.

rolandleth:
Oh, I didn't think of those scenarios. I used to use this with a dynamic-results-list, where it searched and I'd pick what I want from the results. When I tested it now, it was for something obvious and unique 🤦🏻‍♂️

What about only leaving process id, since that's the safest?

Not sure how we'd go for suggestions, as there isn't a place where to display them (makes no sense to make this script `fullOutput` when it succeeds).

grzegorzkrukowski:
If we make script `compact` there is a small Toast message - it can be `success` and `failure` - there is limit for text there though - so failing script and suggesting process there would be an option - but it may not be best experience.

I think we can use `name` as parameter and as long as there is only single process found - then it makes sense to kill it - probably most of the time there will be single one. Otherwise if there are more processes fitting this name I think it's safer to fail and ask user form more explicit name or PID.

rolandleth:
What do you think of the current approach? Tried to implement the best of both worlds.

rolandleth:
> I think that it would be great to first check for full match (caseInsensitive match would be great) and only if there are no applications having exact match, try to search using localizedCaseInsensitiveContains. That will allow to kill Finder by name. It won't work for Finde or Find - but at least it will be doable without digging out the PID by just typing full process name.
Makes sense 👍🏻

> Another minor issue is that message displayed at the end containing all application is quite unreadable - it is way too long and only visible for 2 seconds. Maybe just making it shorter and displaying count of matched applications would be enough to make it readable within time it is displayed.
I thought long about this and tinkered with various options, but I think it's better UX to see what the results are, even if for a short duration (since you can instantly run it again). Think of a non-tech person: telling them there are 4 results won't help them much, as they might not be able to find out which one they want, or not know how to find its process id. It's an easy change to make it show "Found 4 results", just that I think it's even less friendlier.

grzegorzkrukowski:
Yes, currently we cannot make too much to improve UX - having a filtered list of filtered processes to pick from of course will give best user experience. For now we can stick with Toast message as it is - let's just fix the full match use-case and we can merge it.

rolandleth:
Done 👍🏻

PR #293, opened 2021-03-10
🌉 Update preferences.png

Pondorasti:
## Description

Update the preferences image with a screenshot from V1.9

## Type of change

- [x] Documentation update

## Checklist


Pondorasti:
Screenshot 2021-03-09 at 17 27 45

I also feel that bullet no. 4 is outdated. Maybe the following statement would work better?

* Select the folder that contains your Script Commands (ex: [_enabled-commands](https://github.com/raycast/script-commands/tree/master/_enabled-commands))

grzegorzkrukowski:
Good catch with this outdated image - thanks.

About no. 4 - I think we should stick to naming convention used inside Raycast itself more and use `directory` instead of `folder` for consistency.

Pondorasti:
In that case should we do something like this?

1. Open the Extension tab in the Raycast preferences 
2. Select the Script Commands extension or use the keyboard shortcut `⌘ + N`
3. Click `Add Directories` or `Add Script Directory`
4. Select the directory that contains your Script Commands (ex: [_enabled-commands](https://github.com/raycast/script-commands/tree/master/_enabled-commands))

grzegorzkrukowski:
> In that case should we do something like this?
> 
> 1. Open the Extension tab in the Raycast preferences
> 2. Select the Script Commands extension or use the keyboard shortcut `⌘ + N`
> 3. Click `Add Directories` or `Add Script Directory`
Let's make here only `Add Script Directories` - actually now when I look at it it feels a bit inconsistent inside the application itself. We will make both buttons to be `Add Script Directories` in both places for the next version to improve :)

> 4. Select the directory that contains your Script Commands (ex: [_enabled-commands](https://github.com/raycast/script-commands/tree/master/_enabled-commands))

Pondorasti:
Sounds good with me!

grzegorzkrukowski:
After short internal discussion the final suggestion will be to have description for `script command -> add directories` in 1,2,3,4 steps and then mentioning `You can also use add menu by pressing the “+” button or using shortcut Cmd+N` - I can update that part - will do the merge with image and update our documentation.

PR #296, opened 2021-03-10
New Script: SideNotes: Add script to create a new note

MarcelBochtler:
## Description

This script creates a new [SideNotes](https://www.apptorium.com/sidenotes) note in the given folder.
If no folder is specified the note will be created in the open folder or in the first folder as a fallback.

## Type of change

- [x] New script command

## Screenshot

SideNotes new script creation

## Checklist


grzegorzkrukowski (changes requested):
Hey @MarcelBochtler - thanks for your first contribution.

I tried to run the script, but it seems it has a Syntax Error problem: `Expected “then”, etc. but found property.`

commands/apps/sidenotes/sidenotes-create-note.applescript
```diff
@@ -0,0 +1,58 @@
...
+	end if
+
+	tell application "SideNotes"
+		set note_content to item 1 of argv
+		set folder_name to item 2 of argv
+
+		set folder_index to my index_of_folder(folder_name, folders)
+
+		if folder_index is -1 then
+			if current folder is missing value then
```
grzegorzkrukowski, line 35: `Expected “then”, etc. but found property.`

MarcelBochtler: Hi @grzegorzkrukowski, that is strange, as it works for me.
I'll try to repoduce and fix the issue.

fe9lix:
Hi! Any updates here? (Draft status...)

MarcelBochtler:
> Hi! Any updates here? (Draft status...)

I'm waiting to receive my new MacBook and hope to be able to reproduce the mentioned issue then.
Afterwards I will update the PR.

commands/apps/sidenotes/sidenotes-create-note.applescript
MarcelBochtler: @grzegorzkrukowski Unfortunately I'm not able to reproduce your issue. It works for me on Catalina 10.15.7 (can't update right now)

grzegorzkrukowski: We tried it on the second machine with Catalina 10.15.7 - and it fails on the same line.

MarcelBochtler: How are you calling the script?
What parameters are you giving?
I'm also testing on Catalina 10.15.7. Thanks for your patience.

grzegorzkrukowski: I am running it in Raycast. I also tried opening it in default apple script editor and run it from there - it fails on the same line :(

grzegorzkrukowski: Here is preview from editor:

image

grzegorzkrukowski: It seems that the reason was I didn't have `SideNotes` installed on my machine and this was causing compilation errors. After installing SideNotes it works. 

I wonder if we can do something about that case - I was expecting to get the message that I need to install it - I see you are handling that one above, but AppleScript is failing to even run for this case.

grzegorzkrukowski:
Another problem I found after installing SideNote is that it doesn't work if you don't have folders:
CleanShot 2021-04-12 at 14 01 38@2x
CleanShot 2021-04-12 at 14 01 47@2x

commands/apps/sidenotes/sidenotes-create-note.applescript
MarcelBochtler: `current folder` is taken from SideNote's library. That's why the script won't compile when it's not installed.
The check is done the same way for Things which would then also fail if the application is not installed:
https://github.com/raycast/script-commands/blob/50bd467aa8938ade0a27f1a4eec24a934a41efff/commands/apps/things/things-current-todo.applescript#L18-L21

grzegorzkrukowski: Ok fair enough - we can leave it as it is - I didn't find a way to prevent it either. Resolving this issue, leaving one more on the main discussion. Sorry for my lack of AppleScript experience, I was not aware that's gonna fail so hard without app installed :)

grzegorzkrukowski (changes requested):
The script is still failing to create a note when there is no folder created.

We should create a `New Folder` for that case probably or at least show a readable error for a user.

MarcelBochtler:
A new folder is now created if no folders exist.
However: If the user mistypes the folder's name the script falls back to add the note to the first folder.
Is this ok? Or should the script create the "mistyped" folder name and add the note in there?

grzegorzkrukowski:
> A new folder is now created if no folders exist.
> However: If the user mistypes the folder's name the script falls back to add the note to the first folder.
> Is this ok? Or should the script create the "mistyped" folder name and add the note in there?

I think if the user provides a "Folder Name" then it should be created if it doesn't exist.
This way we will provide more functionality - a user can create a Folder directly via script if needed.
If the folder name is not provided we can then simply use the first one from the list as you have it now.
I suggest to ignore miss-typing use case for now.

MarcelBochtler:
> I think if the user provides a "Folder Name" then it should be created if it doesn't exist.

Yes, that makes sense.
Changed the script to do that.

grzegorzkrukowski (approved):
Looks good now - thanks for your hard work.

PR #300, opened 2021-03-12
Add scripts to translate text in DeepL for Mac and the DeepL website

jonohewitt:
## Description

This script takes a command text argument, opens the DeepL for Mac desktop application if it is not already open, and pastes the text into the input text area - overwriting anything that might already exist there.

It could be helpful for anyone regularly checking how to translate a few words or short phrases.

## Type of change

- [x] New script command

## Screenshot

https://user-images.githubusercontent.com/47791294/111003484-84fe1600-8387-11eb-9ff0-9422c82f8f32.mp4

Requires the DeepL for Mac app to be installed. The keys used for select all and paste will need to be configured if they aren't cmd+`a` and cmd + `v`.  The delay to wait for the app to open might also need to be configured on slower macs.

## Checklist


imthath-m:
Hey @jonohewitt, Thanks for the contribution. 

A few thoughts and suggestions.

1. In DeepL, have you set french as the default language to translate into? Is it possible to add an optional argument for the target language and select it while opening the app?
2. You can use the clipboard in a different way. If there is some text already in the clipboard, you can directly open DeepL and translate it. I think that will be a good use case to solve for this type of integration with DeepL. You might need to add it as another script if you would like to.

jonohewitt:
Thanks @imthath-m!

On point 1:

The DeepL app seems to remember the most recently selected language to translate into, and will keep using it until you manually choose a different language. For the inputs it autodetects the language each time.

In terms of selecting the input language within the script command - I can't think of a method that wouldn't break on non-english language systems, since the language selector drop down button wouldn't be selectable with e.g  `UI Element "Into French" of group 1...`

This might have to be a limitation of the script unless another way to select the language dropdown menu element is found. It should be straightforward to do with the web interface though - where the URL is structured like:

[https://www.deepl.com/translator#en/fr/Example%20sentence](https://www.deepl.com/translator#en/fr/Example%20sentence)

I'll try writing an additional web version script for that. 

On point 2:

That's a really cool idea to use the clipboard as an input too. I've adapted the script to make the input argument optional and use the clipboard if nothing has been entered. I've also added an option within the script to copy the translation to the clipboard afterwards.

imthath-m:
Thanks @jonohewitt , for the detailed explanation and implementing the suggestion. 

What do you think about making `copyResultToClipboard` also an optional argument? I'd prefer that it defaults to `true` but if you prefer `false` that is also fine, because users can change that once in the script if they wish to. 

Generally, we can think of making values which users might want to tweak often as arguments rather than having them inside the script. In this case, users might want to change `copyResultToClipboard` often but most likely not `delayTime`.

jonohewitt:
No problem!

We could do - I left the `copyResultToClipboard` out of the args because the DeepL app makes a copy button easily available for occasional use, and I didn't want to add any clutter in the Raycast interface for those who wouldn't use it. What would be the Raycast recommended implementation of a boolean argument - just another text box returning true if anything is typed inside it?

For the default value - I thought it would be safer to make it opt in so that users aren't confused by unexpectedly replacing the clipboard if they didn't read the script thoroughly. It might not be that helpful anyway for users who interact with the DeepL app to customise their initial translation result phrases / switching certain words - as they have to re-copy the translation manually once they're happy with it.

jonohewitt:
Just realised I've put the newer web search script in the apps folder - should I split the two scripts into `Apps` and `Web searches` or perhaps move both into another folder e.g `Communication`, or even `Conversions`?

imthath-m:
I think Apps and Web Searches would be good.

unnamedd:
`Apps` is the best place to hold your Deepl Script Commands @jonohewitt, but you already did that!

jonohewitt:
Okay! Let me know if there's any other changes you'd like me to make - I don't think there's anything left I want to add for now.

PR #303, opened 2021-03-14
Google translate script

rolandleth:
## Description

Performs a Google Translate search and returns the result(s).

## Type of change

- [x] New script command

## Screenshot

2021-03-14 at 10 17 10
custom tl
default-result
custom custom
custom-result

## Checklist


imthath-m:
Hi @rolandleth , Thanks for the contribution.

This will be a welcome addition but we feel it is possible to make it more easily accessible by using 3 query instead of just one. The string you want to translate, source and target languages. 

Ideally, the language params can be optional and you can set some default language. 

This way, you need not modify the script if you want to translate to a different language just once.

rolandleth:
Sure, that makes sense. I used to use it with Alfred, in a search-as-you-type manner, but for Raycast it does make sense to have optional params.

Changes done.

commands/web-searches/google-translate.template.rb
```diff
@@ -0,0 +1,47 @@
...
+# @raycast.schemaVersion 1
+# @raycast.title Google Translate
+# @raycast.mode compact
+# @raycast.packageName Web Searches
+
+# Optional parameters:
+# @raycast.icon 🌍
+# @raycast.argument1 { "type": "text", "placeholder": "Query" }
+# @raycast.argument2 { "type": "text", "placeholder": "from", "optional": true }
+# @raycast.argument3 { "type": "text", "placeholder": "to", "optional": true }
```
imthath-m, line 13:
```suggestion
# @raycast.argument2 { "type": "text", "placeholder": "From", "optional": true }
# @raycast.argument3 { "type": "text", "placeholder": "To", "optional": true }
```

PR #319, opened 2021-03-21
Add scripts to change casing

robertcoopercode:
## Description

Adds 6 scripts to change the casing of text:

- uppercase
- lowercase
- kebab case
- camel case
- title case
- snake case

Essentially just a port of https://github.com/gillibrand/alfred-change-case

## Type of change

- [x] New script command

## Screenshot

CleanShot 2021-03-20 at 19 56 55

Requires python, but should be installed by default on Mac OS

## Checklist


fe9lix (changes requested):
Thanks for your contribution! Could you update the packageName to sth. that better reflects the category, e.g. "Conversions"? (see other conversion scripts...)

robertcoopercode:
> Thanks for your contribution! Could you update the packageName to sth. that better reflects the category, e.g. "Conversions"? (see other conversion scripts...)

Can you explain what `sth` means?

fe9lix:
Sorry: sth. = something

fe9lix:
Regarding the shared code: Would you mind keeping all required code in each of the individual scripts? The reason is that we'd like to keep the scripts self-contained as much as possible (easier sharing). Some duplication is okay.

robertcoopercode:
Updated

robertcoopercode:
I just noticed there is aa script that covers uppercase and lowercase and title case: https://github.com/raycast/script-commands/blob/master/commands/developer-utils/transform-case.sh

The title case result is not as elegant, but the uppercase and lowercase look to work well and are in bash.

fe9lix (approved):

PR #329, opened 2021-03-23
unicode superscript

adam-zethraeus:
## Description

Converts text to fake unicode ˢᵘᵖᵉʳˢᶜʳᶦᵖᵗ.

## Type of change

- [x] New script command

Only default macOS dependencies.

## Checklist

### DIFF
--- a/commands/communication/text/unicode-superscript.sh
+++ b/commands/communication/text/unicode-superscript.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# @raycast.title superscript
+# @raycast.title Unicode Superscript
 # @raycast.author Adam Zethraeus
 # @raycast.authorURL https://github.com/adam-zethraeus
 # @raycast.description Convert clipboards text to fake unicode superscript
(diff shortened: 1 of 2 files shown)


fe9lix (changes requested):
Thanks for your contribution! A couple of comments:

- I would see the script to better fit in the existing "conversions" folder. In that case you could then update the packageName to "Conversions".
- The placeholder text looks a bit "funky" in Raycast so better remove the superscript there.

May I ask what the concrete use case for that script is?

adam-zethraeus:
> Thanks for your contribution! A couple of comments:
> 
> * I would see the script to better fit in the existing "conversions" folder. In that case you could then update the packageName to "Conversions".
> * The placeholder text looks a bit "funky" in Raycast so better remove the superscript there.

@fe9lix done!

> May I ask what the concrete use case for that script is?

Hmm… It's just a small example of a large field of Useful Textexpander-esqueue Things™️. Things like this are fun when communicating nuance and stuff. ᶦ'ᵐ ᵇᵃˢᶦᶜᵃˡˡʸ ˢᵃʸᶦⁿᵍ ᶦᵗ'ˢ ᵃ ᵗᵒᵗᵃˡˡʸ ᵘˢᵉˡᵉˢˢ ᵗᵒʸ

fe9lix (approved):

PR #333, opened 2021-03-25
Add Base64 encode / decode, md5 hash, sha1 hash

hzb:
## Description

Add some script-commands: Base64 encode / decode, MD5 encrypt, SHA1 encrypt

## Type of change

- [X] New script command

## Screenshot

base64

SHA1 requires `md5sha1sum`, you can install via Homebrew. 

```
brew install md5sha1sum
```

I have wrote it on its description.

## Checklist

### DIFF
--- a/commands/system/base64-decode.sh
+++ b/commands/system/base64-decode.sh
@@ -7,7 +7,7 @@
 # @raycast.packageName System
 #
 # Optional parameters:
-# @raycast.icon 🔐
+# @raycast.icon 🔓
 # @raycast.needsConfirmation false
 # @raycast.argument1 { "type": "text", "placeholder": "text", "optional": false }
 #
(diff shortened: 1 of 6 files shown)


fe9lix (changes requested):
Hi! Some comments:

- We do already have base64 encode and decode scripts, please check: https://github.com/raycast/script-commands/tree/master/commands

- MD5 and SHA are hashing functions so I would remove "Encrypt" from the the title and descriptions

- Both md5 and shamus are built-in macOS functions, so there would be no need for a dependency

hzb:
@fe9lix 

- I saw those scripts and they are good. We each have their own advantages in use.

- Rename to `MD5 Hash` and `SHA Hash`, and the descriptions were also updated.

- Upgrade to use the built-in SHA function of macOS.

hzb (commented):
Checked and WFM

fe9lix (approved):

PR #339, opened 2021-03-30
Support conversion of epoch in milliseconds for epoch-to-human-date.sh

francisfeng:
## Description
Current command of `epoch-to-human-date.sh` only supports Unix time in seconds. This pull request enables conversion from Unix time in milliseconds.

## Type of change
- [x] Improvement of an existing script

## Screenshot
Not a new script.

## Dependencies / Requirements
Not needed.

## Checklist

- [x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

PR #340, opened 2021-03-30
Add new conversion script "Unix Time Reader"

francisfeng:
## Description

Display Human-Readable Date from Unix Time in Pasteboard.

There is aleady a `epoch-to-human-date.sh` in this repository and I opened a pull request #339  about it just earlier today.

But Unix Time Reader is so different that I decided to make it as a new script. Specifically, it does not require any arguments, displays result in `compact` mode instead of the fleeting `silent` mode and does not copy the result to the clipboard. It also handles error.

## Type of change

- [x] New script command

## Screenshot

image

## Checklist

### DIFF
--- a/commands/conversions/unix-time-reader.sh
+++ b/commands/conversions/unix-time-reader.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# @raycast.schemaVersion 1
+# @raycast.title Unix Time Reader
+# @raycast.mode compact
+# @raycast.packageName Conversions
+#
+# @raycast.icon 🕰
+# @raycast.needsConfirmation false
+#
+# Documentation:
+# @raycast.description Display Human-Readable Date from Unix Time in Pasteboard
+# @raycast.author Francis Feng
+# @raycast.authorURL https://github.com/francisfeng
+
+unixTime=$(pbpaste)
+size=${#unixTime}
+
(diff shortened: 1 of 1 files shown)


commands/conversions/unix-time-reader.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# @raycast.schemaVersion 1
+# @raycast.title Unix Time Reader
```
mattisssa, line 4: Let's add _" From Clipboard"_ to the title to stay consistent with commands that reading data from the clipboard.

commands/conversions/unix-time-reader.sh
```diff
@@ -0,0 +1,30 @@
...
+size=${#unixTime}
+
+if [[ $size == "10" ]]
+then
+        readable=$(echo `date -r $unixTime "+%F %T"`)
+elif [[ $size == "13" ]]
+then
+        readable=$(echo `date -r $(($unixTime/1000)) "+%F %T"`)
+else
+	echo "Not Unix Time in Pasteboard"
```
mattisssa, line 26: Let's rename the output here to something like "Unix Time is not found"

mattisssa (approved):
Hey! Everything looks good, here are a few comments before the merge

PR #349, opened 2021-04-07
New Script: Toggle Do not disturb

exodusanto:
## Description

Toggle system "Do not disturb", works for Big Sur

## Type of change

- [x] New script command

## Screenshot

Command

Brew package
[calm-notifications](https://github.com/vitorgalvao/tiny-scripts/blob/master/calm-notifications)

## Checklist

### DIFF
--- a/commands/system/toggle-do-not-disturb.sh
+++ b/commands/system/toggle-do-not-disturb.sh
@@ -5,7 +5,7 @@
 
 # Required parameters:
 # @raycast.schemaVersion 1
-# @raycast.title Toggle Do not disturb
+# @raycast.title Toggle Do Not Disturb
 # @raycast.mode compact
 #
 # Optional parameters:
@@ -13,7 +13,7 @@
 # @raycast.packageName System
 #
 # @Documentation:
-# @raycast.description Toggle Do not disturb
+# @raycast.description Toggle Do Not Disturb
 # @raycast.author Antonio Dal Sie
 # @raycast.authorURL https://github.com/exodusanto
 
@@ -22,8 +22,8 @@ then
   calm-notifications off
   # Wait 2s for status bar apply
   sleep 2
-  echo "Do not disturb off"
+  echo "Do Not Disturb off"
 else
   calm-notifications on
-  echo "Do not disturb on"
+  echo "Do Not Disturb on"
 fi
--- a/commands/system/toggle-do-not-disturb.sh
+++ b/commands/system/toggle-do-not-disturb.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Dependency: This script requires `calm-notifications` cli installed: https://github.com/vitorgalvao/tiny-scripts/blob/master/calm-notifications
-# Install via homebrew: `brew install calm-notifications`
+# Install via homebrew: `brew install vitorgalvao/tiny-scripts/calm-notifications`
 
 # Required parameters:
 # @raycast.schemaVersion 1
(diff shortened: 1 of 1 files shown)


commands/system/toggle-do-not-disturb.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Dependency: This script requires `calm-notifications` cli installed: https://github.com/vitorgalvao/tiny-scripts/blob/master/calm-notifications
+# Install via homebrew: `brew install calm-notifications`
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Toggle Do not disturb
```
RSO, line 8:
```suggestion
# @raycast.title Toggle Do Not Disturb
```

commands/system/toggle-do-not-disturb.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# Dependency: This script requires `calm-notifications` cli installed: https://github.com/vitorgalvao/tiny-scripts/blob/master/calm-notifications
+# Install via homebrew: `brew install calm-notifications`
```
RSO, line 4: When I try to install this dependency I get the following error:
```
❯ brew install calm-notifications
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
Error: No available formula or cask with the name "calm-notifications".
==> Searching for a previously deleted formula (in the last month)...
Error: No previously deleted formula found.
==> Searching taps on GitHub...
Error: No formulae found in taps.
```

RSO (changes requested):

commands/system/toggle-do-not-disturb.sh
exodusanto: Sorry, my bad.
Package name is `vitorgalvao/tiny-scripts/calm-notifications`

commands/system/toggle-do-not-disturb.sh
```diff
@@ -0,0 +1,29 @@
...
+if [ "on" = $(calm-notifications status) ]
+then
+  calm-notifications off
+  # Wait 2s for status bar apply
+  sleep 2
+  echo "Do Not Disturb off"
+else
+  calm-notifications on
+  echo "Do Not Disturb on"
+fi
```
RSO, line 29: Seems like `calm-notifications` also supports the `toggle` command, wouldn't the following be simpler?

```suggestion
calm-notifications toggle

echo "Do Not Disturb $(calm-notifications status)"
```

RSO: This even makes me wonder whether we should accept `on/off/toggle/status` as parameter and pass it along 🤔

exodusanto: Original implementation was `calm-notifications toggle`. In the `off` case the DND Icon on status bar change opacity and disappear after some seconds.
That code is to wait some seconds in that case to prevent double toggle request.

But maybe is useless in most cases 😅

RSO: Ah, that makes sense I guess! I'm fine with leaving it at is for now, and see if people like it this way. We can always accept future PRs if need be :-)

RSO (approved):

commands/system/toggle-do-not-disturb.sh
exodusanto: The parameter i see that `default` attribute is not implemented yet, and in my idea the script is a shortcut to toggle. This can be an enhancement with that option in the future 🤗

RSO: I was thinking that if you make the parameter optional you can default to toggling in Bash

PR #351, opened 2021-04-08
Refactor Script: Do Not Disturb

exodusanto:
## Description

This is a possibile enhancement using `optional` attribute.

Commands:
- toggle (default)
- off
- on
- status

The toggle script of #349 will be removed.

## Type of change

## Checklist


commands/system/do-not-disturb.sh
```diff
@@ -0,0 +1,55 @@
...
+# @raycast.authorURL https://github.com/exodusanto
+
+echo_status() {
+  echo "Do Not Disturb $(calm-notifications status)"
+}
+
+off() {
+  calm-notifications off
+  # Wait 2s for status bar apply
+  sleep 2
```
RSO, line 28: Would there be a problem if we simply *always* sleep 2 seconds after toggling? That way we can greatly simplify the code to look something like this:

```bash
if [[ "$1" != "on" && "$1" != "off" && "$1" != "toggle" ]]; then
  echo "Unsupported parameter value: $1 (must be [on/off/toggle])"
  exit 1
fi

calm-notifications $1
sleep 2
echo "Do Not Disturb $(calm-notifications status)"
```

exodusanto: Sure!

RSO (approved):
Thank you for this contribution and the remote pairing session!

PR #363, opened 2021-04-16
Get and set audio devices

rolandleth:
## Description

1. A script that gets a list of all audio devices and stars the currently selected one.
2. A script that sets the selected input or output device, based on the name.

## Type of change

- [x] New script command

## Screenshot

get-input
get-result
set-input-1
set-input-2
set-result

## Checklist


commands/system/audio/set-audio-device.swift
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		return name as String?
+	}
+
+}
+
+let arguments = Array(CommandLine.arguments.dropFirst())
+let query = arguments.first!
+let shouldChangeInput = ["yes", "y", "true"].contains(arguments.last ?? "")
```
grzegorzkrukowski, line 83: using `.last` here makes it easy to have optional here, but on the other hand, it will be safer to make it explicitly get parameter by index in case we add more parameters in the future.
Maybe using a combination of `removeFirst` and `.first`?

commands/system/audio/get-audio-devices.swift
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/swift
+
+// Required parameters:
+// @raycast.schemaVersion 1
+// @raycast.title Audio devices
+// @raycast.mode fullOutput
+
+// Optional parameters:
+// @raycast.icon 🎶
+// @raycast.packageName System
```
grzegorzkrukowski, line 10: Should we make packageName as `Audio` maybe and all other scripts if those are already reorganized into subfolder?

grzegorzkrukowski (changes requested):

commands/system/audio/set-audio-device.swift
rolandleth: I find `.first` would be confusing if you don't notice the `.removeFirst` right away, though 🤔

If we go this route of "what if we add more params in the future", we can also go the "what if we reorder them in the future", then there's no reliable way to handle params anymore 🙃

What about `let shouldChangeInput = arguments.count >= 2 && ["yes", "y", "true"].contains(arguments[1])`?

grzegorzkrukowski: True. Just last is also a bit confusing at first, so lets go with this count option to make it more readable and straightforward

rolandleth: Done 👍

grzegorzkrukowski:
@rolandleth thanks for the late Friday changes! scripts looking good

PR #373, opened 2021-04-23
replace do-not-disturb scripts with one without dependencies

adam-zethraeus:
## Description

This replaces the previous Do Not Disturb toggling scripts that required external dependencies with one which does not.

## Type of change

- [ x ] Improvement of an existing script

## Screenshot

CleanShot 2021-04-22 at 18 10 10@2x

Only that the user grant system accessibility permissions

## Checklist

- [ x ] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)


imthath-m:
Hey Adam, Good attempt in removing the dependency. 

But it doesn't work in all cases. If I have already set a custom DND schedule in System Preferences and then use this to override that for some time, I get the toast that notifications enabled but nothing happens.

adam-zethraeus:
(I got this working to the point that it signals if you're scheduled and doesn't mess with it, so if you want that LMK.)

PR #392, opened 2021-05-01
Add "Slack Good Morning in #general"

farisaziz12:
## Description

Script command that uses AppleScript to send a random good morning message (e.g. "Good Morning all!") to the `#general` chat. There are 3 optional arguments:
1. `channel` send the random good morning text to another specified channel (e.g. `#random`)
2. `message` send another message besides the random good morning message
3. `setActive` takes a `y/n` argument to optionally set yourself as active when sending the good morning text

**Note:** setting status as active only works when inactive. If the user is already active then the message is still sent (hence it being placed before the activation command) and Slackbot will throw an error only visible to you that it's an invalid command.

## Type of change

- [x] New script command

N/A

## Checklist


commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.author Faris Aziz
+# @raycast.authorURL https://github.com/farisaziz12
+# @raycast.schemaVersion 1
+# @raycast.title Slack Good Morning in #general
+# @raycast.mode silent
+# @raycast.packageName Slack Channel Good Morning
```
unnamedd, line 9:
```suggestion
# @raycast.packageName Slack
```

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/osascript
```
unnamedd, line 1: The file should be renamed to `send-slack-message.applescript` or `slack-send-message.applescript`.

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
...
+
+# Required parameters:
+# @raycast.author Faris Aziz
+# @raycast.authorURL https://github.com/farisaziz12
+# @raycast.schemaVersion 1
+# @raycast.title Slack Good Morning in #general
```
unnamedd, line 7: As you might expect the user to define (or not) which channel it is going to send the message (which can be “Good Morning” or not), then IMO we should remove `#general` of the title and use `Channel`:
```suggestion
# @raycast.title Send Message in Channel
```

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
...
+# @raycast.mode silent
+# @raycast.packageName Slack Channel Good Morning
+# @raycast.description This script sends a random good morning text to the #general channel and sets active status (can be changed using args)
```
unnamedd, line 10:
```suggestion
# @raycast.description This script posts a message text to the a slack channel and sets active status (can be changed using args)
```

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
...
+# @raycast.needsConfirmation true
+# @raycast.argument1 { "type": "text", "placeholder": "channel (default: #general)", "optional": true, }
```
unnamedd, line 12:
```suggestion
# @raycast.argument1 { "type": "text", "placeholder": "Channel (default: #general)", "optional": true, }
```

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
...
+# @raycast.argument2 { "type": "text", "placeholder": "message (default: good morning)", "optional": true, }
```
unnamedd, line 13:
```suggestion
# @raycast.argument2 { "type": "text", "placeholder": "Message (default: good morning)", "optional": true, }
```

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
...
+# @raycast.argument3 { "type": "text", "placeholder": "set active? y/n", "optional": true, }
```
unnamedd, line 14:
```suggestion
# @raycast.argument3 { "type": "text", "placeholder": "Set Active? y/n", "optional": true, }
```

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
...
+-- if no message argument set random good morning message as default
+    if message = "" then
+        set messageToSend to randomGoodMorningText
+    else
+        set messageToSend to message
+    end if
+
+    openChannel(channelToSendTo)
+    -- setting to active will not work if already active
+    if setActive = "y" then
```
unnamedd, line 96: You should support an uppercased value, in case of a user’s mistake, your Script Command won’t stop to work.

commands/communication/slack/send-channel-good-morning.applescript
```diff
@@ -0,0 +1,107 @@
...
+        slashCommand("Set yourself as active")
+        pressReturn()
+        log messageToSend & " sent To " & channelToSendTo &" and status set to active!"
+    else if setActive = "n" or setActive = "" -- default is no
+        sendMessage(messageToSend)
+        log messageToSend & " sent To " & channelToSendTo &"!"
+    else
+        log "Invalid argument for set active. Specify setting active using 'y' or 'n'"
+    end if
+end run
```
unnamedd, line 107: Do you need all those `delays` in the methods `openChannel`, `sendMessage`, `slashCommand` and `pressReturn`? I do think that half second is a bit much and calling method after method, it should take at least 4.5 seconds if I counted correctly to finish the execution of your Script Command successfully.

unnamedd (changes requested):
Hello @farisaziz12,
please, take a look into my comments and after address them, please, let us know!

Thanks for your contribution!

commands/communication/slack/send-channel-good-morning.applescript
farisaziz12: Makes sense! Thanks for the suggestion

commands/communication/slack/send-channel-good-morning.applescript
farisaziz12:
```suggestion
# @raycast.description This script posts a message text to a slack channel and sets active status (defaults to random good morning message in #general)
```
@unnamedd  how about keeping the main description about general usage of the script and in brackets mentioning default behaviour? This way the user does not need to dig through and understand the code to understand this. 😄

unnamedd: Sounds good to me if you mention the default value. Explicit information is always better than an implicit ones.

commands/communication/slack/send-channel-good-morning.applescript
farisaziz12: @unnamedd Thank you so much for taking the time to make your suggestions. I have actioned all of them and brought down the execution time to maximum 4 seconds. Any lower than this and the script becomes unreliable. Reducing execution time to 3.5 seconds yields issues such as this:
Screenshot 2021-05-03 at 12 56 29

unnamedd: Nice, thanks for take a time addressing my concerns. If this is the minimum we can have, then let's go. 

Maybe a nice thing to do as an improvement here, is to use Slack API in the future.

farisaziz12: I would have used the slack API but wanted to reduce the hassle of getting a token to verify requests. This way you don't have to create and install an app to Slack (which may not always be an option for professional workspaces).

dehesa:
Thanks for this @farisaziz12

PR #393, opened 2021-05-02
Bitwarden: Add script that autotypes the first result

MarcelBochtler:
## Description

This script autotypes the username and password of the first found result in Bitwarden.
It requires the username field to be selected and the password field to be reachable with one TAB key press.

## Type of change

- [x] New script command

## Screenshot

https://user-images.githubusercontent.com/4022103/116819953-d9956480-ab72-11eb-9590-f15289c74cdc.mov

Dependencies:
1. The Bitwarden CLI: https://bitwarden.com/help/article/cli/
2. The `jq` utility: https://stedolan.github.io/jq/

## Checklist

### DIFF
--- a/commands/password-managers/bitwarden/autotype-first-result.sh
+++ b/commands/password-managers/bitwarden/autotype-first-result.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# Dependencies:
+#   1. The Bitwarden CLI: https://bitwarden.com/help/article/cli/
+#   2. The `jq` utility: https://stedolan.github.io/jq/
+#
+# Install via homebrew: `brew install bitwarden-cli jq`
+# or via macports: `sudo port install bitwarden-cli jq`
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Autotype First Result
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.packageName Bitwarden
+# @raycast.icon images/bitwarden.png
+# @raycast.argument1 { "type": "text", "placeholder": "Query" }
+
+# Documentation:
+# @raycast.author Marcel Bochtler
+# @raycast.authorURL https://github.com/MarcelBochtler
+# @raycast.description Type the username and password of the first Bitwarden result. Separated by one {TAB}.
+
+notFound() {
+  echo "The query '$1' did not return a result."
+  exit 1
+}
+
+if ! command -v bw &> /dev/null; then
+  echo "The Bitwarden CLI is not installed."
+  exit 1
+elif ! command -v jq &> /dev/null; then
+  echo "The jq utility is not installed."
+  exit 1
+fi
+
+token=$(security find-generic-password -a ${USER} -s raycast-bitwarden -w 2> /dev/null)
+token_status=$?
+
+session=""
+if [ $token_status -eq 0 ]; then
+  session="--session $token"
+fi
+
+bw unlock --check $session > /dev/null 2>&1
+unlocked_status=$?
+
+if [ $unlocked_status -ne 0 ]; then
+  echo "Vault is locked!"
+  exit 1
+fi
+
+item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { name: .name, username: .login.username, password: .login.password }")
+name=$(echo $item | jq --exit-status ".name") || notFoundError
+username=$(echo $item | jq --raw-output --exit-status ".username") || notFoundError
+password=$(echo $item | jq --raw-output --exit-status ".password") || notFoundError
+
+# TODO: This requires that the username text field to be selected.
+# It might be possible to check if a text field is selected before starting to type.
+osascript <<EOF
+    tell application "System Events"
+        keystroke "$username"
+        key code 48
+        keystroke "$password"
+    end tell
+EOF
+
+unset item
+unset username
+unset password
+
+echo "Autotyped username and password for $name."


commands/password-managers/bitwarden/autotype-first-result.sh
```diff
@@ -0,0 +1,73 @@
...
+        key code 48
+        keystroke "$password"
+    end tell
+EOF
+
+unset item
+unset username
+unset password
+
+echo "Autotyped username and password for $name."
```
unnamedd, line 73: As the Script Command has the mode `silent`, none of the `echo`s above should be shown to the user.

unnamedd (commented):
Very interesting of what you have done, also I’d suggest an improvement to your Script Command: what about if you, using the Apple Script, get the current selected tab in the current opened window, getting the URL of the website and sending to Bitwarden, instead of type the name of the website? Making everything more automatic? 

What do you think @MarcelBochtler?

commands/password-managers/bitwarden/autotype-first-result.sh
MarcelBochtler: This is not correct. The `echo`s are printed as a "toast message"

Screenshot 2021-05-04 at 18 06 28

MarcelBochtler:
> what about if you, using the Apple Script, get the current selected tab in the current opened window, getting the URL of the website and sending to Bitwarden

I like the idea.
But this would require a separate logic for every bowser. e.g.:
```applescript
property theURL : ""

tell application "Chromium"
	set theURL to URL of active tab of window 1
end tell

tell application "Safari"
	set theURL to URL of current tab of window 1
end tell
```
(note active != current)

I also think that the current approach would still be useful for "non-browser" applications.

So the full functionality would be:
If a query is entered. Use the query.
Otherwise check which browser is being used, get the URL, search the URL in `bw` and autotype.

I'll try to add this functionality, but I won't be able to cover each and every browser.

commands/password-managers/bitwarden/autotype-first-result.sh
unnamedd: Yes, you are right. I forgot about that toast. The problem to use it, is the time and where it is being shown to the user, it's really easy to miss the message.

unnamedd:
> So the full functionality would be:
> If a query is entered. Use the query.
> Otherwise check which browser is being used, get the URL, search the URL in `bw` and autotype.

And here we are with a nice brainstorm on how the Script Command can be improved! I really love that! 
You don't need to do this right now, it's just an idea that you might want to implement later on. I'm completely sure that the functionality suggested by you, would be a very nice improvement for this Script Command and a lot of people (who uses Bitwarden) might be amazed by the result!

commands/password-managers/bitwarden/autotype-first-result.sh
MarcelBochtler: Would you suggest to remove the messages?

unnamedd: Not at all, I'm just concerned for the user not being able (or quick enough) to see the message and understand what happened. 

@PitNikola and @thomaspaulmann what would you suggest for the situation we are discussing here? I'm a bit confused based in the needs of the Script Command what should do to a better presentation of possible errors. 

The main intention of the Script Command is to avoid to keep Raycast opened, because the user will be asking for the password, hence it won't want to have Raycast blocking his vision of the page after the request of password, in the other hand, for the user who has a huge monitor screen (like me), it may not find out the message in case of error. For the first point, `silent` mode is the best one, but to show errors, the `compact` mode is better.

I do think your opinions here will be highly appreciated.

commands/password-managers/bitwarden/autotype-first-result.sh
```diff
@@ -0,0 +1,73 @@
...
+username=$(echo $item | jq --raw-output --exit-status ".username") || notFoundError
+password=$(echo $item | jq --raw-output --exit-status ".password") || notFoundError
+
+# TODO: This requires that the username text field to be selected.
+# It might be possible to check if a text field is selected before starting to type.
+osascript <<EOF
+    tell application "System Events"
+        keystroke "$username"
+        key code 48
+        keystroke "$password"
```
PSalant726, line 65: This makes several assumptions about the structure of the selected form, and for that reason it is extremely insecure.

PSalant726 (changes requested):
I'm thrilled that you chose to contribute, and I'm grateful that you made an effort to assimilate this script with the others (consistency is important!), but I would strongly recommend closing this PR until there is native/secure support for autofill via the `bw` CLI.

While I understand that this script may appear to solve for a minor pain point with this package (namely, an "autofill" feature does not exist), I have significant concerns about this implementation. As written, this exposes the user to a common vulnerability: a hidden form field could be automatically populated without the user's knowledge. This is ultimately why I chose not to implement anything like a `populate.sh` script alongside the others, and instead chose to implement [`copy-first-matching-password.sh`](https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/copy-first-matching-password.sh).

Separately, if you plan to disregard my concerns and add this script anyway, it would be helpful if you could also add it to the documentation in [`bitwarden/README.md`](https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/README.md).

dehesa:
Hi @MarcelBochtler, thank you for taking the time to submit this PR. We had a chat internally about your script and we have decided not to include the script in the repo.

I want to stress how happy we are you made the effort to share this here and I want to encourage you to keep sending PRs with whatever you feel it is useful. However, we do reviews here to vet these kind of gray-zone security issues. Obviously, you are free to keep using it as you see fit or even share this script through your own Github repos.

I also want to thank @unnamedd and @PSalant726 for leaving their input here.

MarcelBochtler:
Thanks for your reviews.
I'm fine with not merging this if you see a security issue.

But tbh I don't see the attack vector here. The script did explicitly not include an {ENTER} key press because the password might have been entered into a wrong text field.
Isn't [KeePass](https://keepass.info/help/base/autotype.html) doing exactly the same?
Also the users would be using this functionality where they want to login, and therefore trust this website / application.

PSalant726:
@MarcelBochtler Reputable password managers (including Keepass) use more complex means to determine where to autofill credentials, as well as if/when it's unsafe to do so.

MarcelBochtler:
@PSalant726
> Reputable password managers (including Keepass) use more complex means to determine where to autofill credentials, as well as if/when it's unsafe to do so.

In KeePass you can enter any keystoke sequence that will be entered into any textfield.
E.g. the example in the documentation I linked above:
> By default, the sent keystroke sequence is {USERNAME}{TAB}{PASSWORD}{ENTER}

How does that "determine where to autofill credentials, as well as if/when it's unsafe to do so"?

PSalant726:
Yikes, I made the assumption that Keepass is a reputable password manager...

PR #413, opened 2021-05-12
Add more Espanso commands

es183923:
## Description

Commands to start, stop, enable, disable, and restart espanso.

## Type of change

- [x] New script command

Requires espanso cli (https://espanso.org/install/)

## Checklist


nhojb:
Thanks for the PR 😄 

Rather than multiple scripts it would be better if you passed argument(s) from Raycast to a single script. See here:

https://github.com/raycast/script-commands/blob/master/documentation/ARGUMENTS.md

unnamedd:
> Thanks for the PR 😄
> 
> Rather than multiple scripts it would be better if you passed argument(s) from Raycast to a single script. See here:
> 
> https://github.com/raycast/script-commands/blob/master/documentation/ARGUMENTS.md

Hey John, 
not always a single Script Command is a good choice, in special because the user will have to remember which kind of command it will have write in the argument field, also, we have a lot os Script Commands here which are working with only one command, like: stop, start, enable, disable. It is simple when we are searching for some package name (just one example) and see a list of Script Commands available.

Talking specifically about this pull request, I truly believe the best option is to have multiple Script Commands. What do you think? Did I have any success changing your mind? 😄

nhojb:
Thanks for the feedback and PR, you've convinced us 😄

PR #415, opened 2021-05-15
Update existing Hue scripts and add inline summary scripts

jonohewitt:
### Existing bash scripts rewritten into Javascript using the [zx](https://github.com/google/zx) npm package

I wanted the scripts for setting rooms of lights and individual lights to be able to react to existing state data by parsing a response from the hue bridge. Unfortunately I don't know enough about native bash tools to be able to parse the JSON response without a tool like jq, and if I had to ask users to install a package, I would rather ask them to install zx - a node wrapper with some useful built in tools - and write the entire scripts in Javascript instead.

#### Improvements: 
• Toggle lights by fetching current 'on' state and returning the inverse if no arguments are given
• Can use the bulb or room names already assigned to the Hue bridge by default without having to write them into the script manually.
• Use silent mode instead of compact for cleaner experience
• More thorough input testing and error messages
• Inputs checked with switch statements now, allowing easier stacking of multiple inputs the same output
• (Opinion) I expect more people would feel comfortable editing a Javascript template than bash?

#### Disadvantages:
• Now depends on installing the [zx](https://github.com/google/zx) npm package

## New scripts

Added two new scripts for showing inline summaries of which lights / rooms are on and various data such as how many, what brightness level, colour info etc.

## Type of change

- [x] New script command

## Screenshot
The new inline scripts:

Screenshot 2021-05-15 at 13 49 56

• [zx](https://github.com/google/zx) npm package for running the scripts.
• Philips Hue Bridge IP address and API username
• Some scripts require opening the Hue API in a browser and finding ID numbers - steps are explained within the scripts.

## Checklist


nhojb:
Hi Jon, this is a great contribution. Thanks for the PR!

PR #448, opened 2021-05-27
Added script that updates installed scripts from this repo

qeude:
## Description

Script command that updates installed scripts by getting newest version on this git repository.

## Type of change

- [x] New script command

Requiring to specify root scripts directory.
Optionally can specify ignored scripts.

## Checklist


grzegorzkrukowski:
Hey - great idea! Thanks for your contribution.

When I was testing it I was missing information about the results of the script - if anything got updated or everything was up to date. 

Do you think it makes sense to make the successful message a bit more explicit and maybe check how many scripts got updated or if everything was up to date?

Also at the top of the script, it will be good to mention the required setup, you can see example here:
https://github.com/raycast/script-commands/blob/4f1727eaa3e90c02fa711a61a736797a57b18848/commands/communication/slack/clear-slack-status.template.sh#L3-L5

or here:

https://github.com/raycast/script-commands/blob/4f1727eaa3e90c02fa711a61a736797a57b18848/commands/system/toggle-flip-screen.template.applescript#L3-L9

qeude:
Hi, thanks for your feedback !

> Do you think it makes sense to make the successful message a bit more explicit and maybe check how many scripts got updated or if everything was up to date?

This was my idea at the beginning, but since scripts aren't versioned I am not able to determine scripts that needs update.
Adding the number of updated scripts would always return the same result, even if you are updating those two times in a row. That seems a bit weird to have a feedback which doesn't give this much relevant information more than the script worked and everything is up to date as it does currently.

> Also at the top of the script, it will be good to mention the required setup, you can see example here:

Sure I'll do that ASAP 👍🏼

unnamedd:
Hi @qeude, I guess I can help you out if I fix in the Toolkit exactly the generation of the hash which indicate if some script command was updated and which was, but in that case you will have to run recursively in the nodes of the json generated by Toolkit checking the current SHA with the one you will have to save in your end. I guess this kind of improvement may increase a lot the complexity of your Script Command, but it is doable.

qeude:
Hi @unnamedd,
Yeah sure we could give it a try. I doesn't have an exact vision on how it could be done, but I'm not sure it will increase the complexity this much if it's just about SHA comparison ?

If you need help about fixing the Toolkit, feel free to explain the problem so we could maybe figure out what is going on together ;)

unnamedd:
The generation of the SHA in the Toolkit isn't complex to solve, as I wrote it, it's quite simple to explain to you how can we fix this, the only problem I'm facing now is: **time** 😔. Reach me out on Slack and I can guide through the code, as far as I could see, Swift isn't a problem for you, right?

SHA:
Yes, the comparison of two strings won't increase the complexity, this is trivial, but the problem doesn't lie on the comparison but in the source of the original SHA to be compared to the new one. You will need to save the old SHA somewhere, you can run throughout of the JSON nodes saving the SHA into some file to check it later to see if you have any change on the SHA you are checking now, this where I see you may increase the complexity of your Script Command, which suppose to be something simple.

Again, reach me out on Slack to discuss more about these topics, if you want. There, I am @ tholanda.

grzegorzkrukowski:
@qeude let me know if you want to first merge the existing version and then work on improvements over the next iterations. 
From my perspective only blocker was the header information - this is more nice-to-have.

qeude:
@grzegorzkrukowski I did add the require setup in the script.

I think it's a good idea to merge it first then improving it since I will take a look at the Toolkit so I can figure out if I can fix it. I think we don't want PR to keeps open dying. 😄

qeude:
Waiting for #455 to be merged.

grzegorzkrukowski:
@qeude I merged #455 so you can see if that works. I will merge this PR as it is and feel free to open another one with improvements once you figure out Toolkit problems

unnamedd:
@grzegorzkrukowski the #455 wasn't tested yet, I preferred to have the confirmation that everything was working fine before get it merged, but I checked the result (the generation of the extensions.json) and everything is working as expected then, nothing was harmed! Win win!

PR #455, opened 2021-06-03
Fix Toolkit that isn't retrieving the right createdDate and updatedDate for scripts

qeude:
## Description

Currently the Toolkit isn't setting the right dates for scripts in the `extensions.json` file. These dates are retrieved from the git history via the git cli. The github's checkout action have a parameter `fetch-depth` with is set to 1 by default, this involves that git doesn’t get the full git log history. Setting it to 0 as should fix the history problem and then the weird behaviour of the command. :smile:

## Type of change

- [x] Bug fix
- [x] Toolkit change

## Checklist

### DIFF
--- a/.github/workflows/script-commands.yml
+++ b/.github/workflows/script-commands.yml
@@ -2,7 +2,7 @@ name: Set as Executable and Generate Documentation
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   scriptCommands:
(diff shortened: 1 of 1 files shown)


unnamedd:
@qeude please, confirm if the behavior you mentioned in the PR’s description is indeed working in your fork. 
You should see as a result, different dates for each script command in the properties `createdAt` and `updatedAt` inside of the `extensions.json`.

PR #456, opened 2021-06-03
Added handling of versioning using extension.json on the update-script-command

qeude:
## Description

Improving [update-scripts-command](https://github.com/raycast/script-commands/blob/master/commands/system/update-scripts-command.template.py) to handle script versionning using the [file](https://github.com/raycast/script-commands/blob/master/commands/extensions.json) generated by the Toolkit.
It now only updates when needed and the user's feedback about what happened is improved.
See #448 

## Type of change

## Checklist

### DIFF
--- a/commands/system/update-scripts-command.template.py
+++ b/commands/system/update-scripts-command.template.py
@@ -7,8 +7,9 @@
 
 # Required parameters:
 # @raycast.schemaVersion 1
-# @raycast.title Update Community Scripts
-# @raycast.mode inline
+# @raycast.title Update Community Scripts 
+# @raycast.description Updates community scripts to their last available version from the Github repository.
+# @raycast.mode compact
 # @raycast.refreshTime 1d
 
 # Optional parameters:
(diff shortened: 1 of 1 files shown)


commands/system/update-scripts-command.template.py
```diff
@@ -7,8 +7,9 @@
 
 # Required parameters:
 # @raycast.schemaVersion 1
-# @raycast.title Update Community Scripts
-# @raycast.mode inline
+# @raycast.title Update Community Scripts 
+# @raycast.description Updates community scripts to their last available version from the Github repository.
```
unnamedd, line 11:
```suggestion
# @raycast.description Updates community Script Commands to their last available version from the GitHub repository.
```

unnamedd (commented):
Thanks for the improvement @qeude! Nice to see your Script Command making use of the `extensions.json`.

grzegorzkrukowski:
Wow - really quick and nice update.

I am thinking of one last thing: do we need a `.template` file here or could we just use script current directory `./` as default and skip the manual setup?

qeude:
@grzegorzkrukowski I'm not sure it would be a good idea since the scripts isn't located in your scripts root dir for sure that's why I made a ˋ.template`

grzegorzkrukowski:
Ok merging it as it is then !

PR #467, opened 2021-06-09
Add commands for managing Tailscale

rosszurowski:
## Description

This PR adds a few simple shell commands for managing [Tailscale](https://tailscale.com), a networking tool for connecting devices and servers in a peer-to-peer way.

There are four new commands:
- **Connect** which … connects to Tailscale
- **Disconnect** which … disconnects from Tailscale
- **Get IP** is an inline script which gets your current [private Tailscale IP](https://tailscale.com/kb/1033/ip-and-dns-addresses/), which is how devices talk to each other over Tailscale
- **Switch Account** is a fast script for logging out and logging in as a new user

In a future change, I might add an optional parameter to the scripts to allow defining [a set of auth keys](https://tailscale.com/kb/1085/auth-keys/), circumventing an interactive login for the connect and switch account commands.

## Type of change

- [x] New script command

## Screenshot

https://user-images.githubusercontent.com/303731/121373822-63241900-c90d-11eb-805e-a2395825021e.mp4

These commands require [Tailscale to be installed](https://tailscale.com/download). The scripts annotate this at the top, and check for the presence of either the [Mac App Store distribution](https://apps.apple.com/ca/app/tailscale/id1475387142?mt=12), or [the CLI distribution](https://github.com/tailscale/tailscale/wiki/Tailscaled-on-macOS).

## Checklist


commands/apps/tailscale/tailscale-connect.sh
```diff
@@ -0,0 +1,34 @@
...
+# @raycast.packageName Tailscale
+
+# @Documentation:
+# @raycast.description Connects to Tailscale
+# @raycast.author Ross Zurowski
+# @raycast.authorURL https://github.com/rosszurowski
+
+ts=""
+
+if command -v tailscale &> /dev/null; then
```
unnamedd, line 23: As I'm seeing you have the following piece of code duplicated through all of the Script Commands, you may want to add that specific piece of code in a separated file (without add the Raycast meta tags) and then import it in each Script Command to make your verification, you'll just need to call `source name-of-the-check-file.sh` and to use the $ts variable, just need to export it from there.

unnamedd (changes requested):
Hi @rosszurowski, as a final request, I've like to kindly ask you to write a nice and small README to add in this folder, like was done here: https://github.com/raycast/script-commands/tree/master/commands/apps/things

Thanks for you contribution!

commands/apps/tailscale/tailscale-connect.sh
fe9lix: Please also keep in mind that the `jq` dependency is not pre-installed on macOS.

unnamedd: very good point @fe9lix!

rosszurowski: Yeah, this is great feedback. I'll…

- remove or document the `jq` requirement
- add a readme.md to the folder

I'd prefer to not `source check-file.sh` since some folks (or at the very least, I) like to copy over only the scripts they need. Introducing dependencies between the files makes this a little harder.

rosszurowski: Updated the PR with a README.md in the folder, and made the `jq` check optional. If it's there, it'll show which account you're authenticated as, but otherwise, it just says "Connected". The README.md file calls out the extra feature if you have `jq` installed.

I'd consider used some kind of string grepping to duplicate the functionality without `jq`, but I worry about the format of the file changing or being too complex to parse without `jq`, so I think dropping the bonus functionality is the right move.

fe9lix (approved):

PR #470, opened 2021-06-10
Added create gif from video script

qeude:
## Description

Converts a video file to a gif file using [gifski](https://github.com/ImageOptim/gifski), arguments are all optionals, first one is source file path, second on is destination file path, and last one is resulting gif fps number.
By default, it takes the last video file on Desktop. I've done that to easily convert screen videos.
## Type of change

- [x] New script command

## Screenshot

generated_gif

Needs [gifski](https://github.com/ImageOptim/gifski) to work.

## Checklist

### DIFF
--- a/commands/conversions/create-gif-from-video.py
+++ b/commands/conversions/create-gif-from-video.py
@@ -32,6 +32,10 @@ if which("gifski") is None:
     print("gifski is required (https://gif.ski/).")
     exit(1)
 
+if which("ffmpeg") is None:
+    print("ffmpeg is required (https://www.ffmpeg.org/).")
+    exit(1)
+
 base_directory=f"{os.environ['HOME']}/Desktop"
 
 default_gif_filename="generated_gif.gif"
--- a/commands/conversions/create-gif-from-video.py
+++ b/commands/conversions/create-gif-from-video.py
@@ -5,14 +5,14 @@
 # Required parameters:
 # @raycast.schemaVersion 1
 # @raycast.title Create GIF from video
-# @raycast.mode silent
+# @raycast.mode compact
 
 # Optional parameters:
 # @raycast.icon 📹
 # @raycast.argument1 { "type": "text", "placeholder": "Path to the source file (default: take the last screen recording on the Desktop)", "optional": true }
-# @raycast.argument2 { "type": "text", "placeholder": "Path to the destination file (default: generate a file in the same directory as the source file)", "optional": true }
-# @raycast.argument3 { "type": "text", "placeholder": "FPS (between 1 and 60, default: 60)", "optional": true }
-# @raycast.packageName Conversions
+# @raycast.argument2 { "type": "text", "placeholder": "FPS (between 1 and 60, default: 20)", "optional": true }
+# @raycast.argument3 { "type": "text", "placeholder": "Width (default 600)", "optional": true }
+# @raycast.packageName Utilities
 
 # Documentation:
 # @raycast.description Create a GIF from video, by default it takes the last screen record video
@@ -53,9 +53,6 @@ def safe_get(array, index):
         return None
 
 from_file_path=safe_get(sys.argv, 1) or get_last_video_file_path()
-to_file_path=safe_get(sys.argv, 2) or get_default_destination_file_path(from_file_path=from_file_path)
-frame_rate=safe_get(sys.argv, 3) or 30
-
 if not from_file_path:
     print("No video file have been found at the specified path.")
     exit(1)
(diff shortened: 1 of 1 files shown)


commands/conversions/create-gif-from-video.py
```diff
@@ -0,0 +1,95 @@
...
+# Dependency: This script requires `gifski` to be installed: https://gif.ski/
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Create GIF from video
+# @raycast.mode compact
+
+# Optional parameters:
+# @raycast.icon 📹
+# @raycast.argument1 { "type": "text", "placeholder": "Path to the source file (default: take the last screen recording on the Desktop)", "optional": true }
```
unnamedd, line 12: Avoid huge placeholders, they won't be visible anyway.
```suggestion
# @raycast.argument1 { "type": "text", "placeholder": "Source File Path", "optional": true }
```

commands/conversions/create-gif-from-video.py
```diff
@@ -0,0 +1,95 @@
...
+# @raycast.argument2 { "type": "text", "placeholder": "FPS (between 1 and 60, default: 20)", "optional": true }
```
unnamedd, line 13:
```suggestion
# @raycast.argument2 { "type": "text", "placeholder": "FPS (Default: 20)", "optional": true }
```

commands/conversions/create-gif-from-video.py
```diff
@@ -0,0 +1,95 @@
...
+# @raycast.argument3 { "type": "text", "placeholder": "Width (default 600)", "optional": true }
```
unnamedd, line 14:
```suggestion
# @raycast.argument3 { "type": "text", "placeholder": "Width (Default: 600)", "optional": true }
```

commands/conversions/create-gif-from-video.py
```diff
@@ -0,0 +1,95 @@
...
+# @raycast.packageName Utilities
```
unnamedd, line 15:
```suggestion
# @raycast.packageName Conversions
```

commands/conversions/create-gif-from-video.py
```diff
@@ -0,0 +1,95 @@
...
+# @raycast.author Quentin Eude
+# @raycast.authorURL https://github.com/qeude
+
+import sys
+import mimetypes
+from glob import glob
+import os
+from shutil import which
+import subprocess
+from pathlib import Path
```
unnamedd, line 28: A small suggestion just to organize
```suggestion
import mimetypes
import os
import subprocess
import sys

from glob import glob
from pathlib import Path
from shutil import which
```

unnamedd (changes requested):

commands/conversions/create-gif-from-video.py
```diff
@@ -0,0 +1,96 @@
...
+try:
+    int_frame_rate = int(frame_rate)
+except Exception as e:
+    print(f"Frame rate should be valid integer between 1 and 60.")
+    exit(1)
+
+if int_frame_rate <= 60 and int_frame_rate >= 1:
+    mp4_filename=f"{os.path.splitext(os.path.basename(from_file_path))[0]}.mp4"
+    mp4_file_path=os.path.join(os.path.dirname(from_file_path), mp4_filename)
+    os.system(f"ffmpeg -y -loglevel panic -err_detect aggressive -fflags discardcorrupt -i '{from_file_path}' -c:v libx264 -preset slow -crf 18 -c:a copy '{mp4_file_path}'>/dev/null")
```
fe9lix, line 88: Please also check for ffmpeg in the dependency check above.

fe9lix (changes requested):
Hey @qeude ! Just quickly ran the script here locally with an mp4 (screencapture recorded from CleanShot) and ended up in a situation where I got a success message, but the original mp4 file deleted and no gif actually created. I guess this needs some more error handling because it's obviously undesired that my original file was deleted when running into the error.
Not sure what the actual issue is with the gif not being created...

qeude:
Hi @fe9lix ! I've got an idea about why your file has been deleted, it's my bad, I'll figure it out when I'll have some free time.
About the error handling, I'll try to do something but when I did before for `gifski` I wasn't able to redirect output to `/dev/null` while keeping errors.

fe9lix:
Sounds good – I guess showing the error to the user and in that case leaving the original file in place would be an improvement. I'll leave the PR open for now and you can just continue working on it when you have some time.

qeude:
@fe9lix I made the fix about the deletion of the source file. Wasn't able to reproduce the not generated gif file. I would maybe need some help to handle error of commands nicely.

fe9lix:
Hey @qeude ! I ran the ffmpeg and gifski commands directly in the shell and I got an error on gifski complaining that it hasn't been compiled with "video support". So that seems something specific to me setup here. Nevertheless, I think the error should be handled and then shown to the user. Afaik, os.system doesn't allow you to capture the output of the standard streams, so you would need to use a different way of capturing the output of stderr in Python (Popen maybe). After capturing stderr, you can print it and exit with a non-zero code.

qeude:
Hey @fe9lix, was a bit tought to find time to look at it.
I did commit what I think should fix the handling of stderr, since I'm not able to reproduce your problem, could you maybe confirm that it's working pls ?

fe9lix:
Hey @qeude ! Thanks for your updates and sorry for the reviewing delay! I've just retested and see an error message now (video disabled...). I guess we can merge it now since it works for you and I seem to have a weird local setup here. If something else pops up in the future, you can just iterate on it and further improve it. Cheers!

PR #472, opened 2021-06-11
Add genpasswd2 script

everdrone:
## Description

Generates password according to [Password Rules Validation Tool](https://developer.apple.com/password-rules) by Apple

Much like iOS Safari, it generates passwords like `cesblw-xrdgeb-Gjdpi1`.

This script is based on [genpasswd2](https://github.com/everdrone/genpasswd2)

It's bundled with both light and dark icons (from SF Symbols for better integration in the ecosystem)

## Type of change

- [x] New script command

## Screenshot

Screen Shot 2021-06-11 at 14 57 52

python3

## Checklist


commands/developer-utils/genpasswd.py
```diff
@@ -0,0 +1,80 @@
...
+    length: int = 3,
+    set_length: int = 6,
+    numbers: int = 1,
+    uppercase: int = 1,
+    separator: str = '-'
+) -> str:
+    print(numbers)
+    # if numbers + uppercase <= set_length * length:
+    #     print('The amount of numbers plus the capital letters must be less than or equal to the total amount of characters')
+    #     sys.exit(1)
```
unnamedd, line 38: Are the lines above intended to be used? If not, I'd highly suggest to remove them.

commands/developer-utils/genpasswd.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Generate Password
+# @raycast.mode silent
+# @raycast.packageName Security
+
```
unnamedd, line 8:
```suggestion
# @raycast.argument1 {"type": "text", "placeholder": "Password Length", "optional": true}
```

commands/developer-utils/genpasswd.py
```diff
@@ -0,0 +1,80 @@
...
+    # split and join with separators
+    sets = wrap(all, set_length)
+    result = separator.join(sets)
+
+    subprocess.run("pbcopy", universal_newlines=True, input=result)
+
+    return result
+
+
+print(superset())
```
unnamedd, line 80: Would be nice if you make use of the metatag @raycast.argument and let the choice of the length to the user.
```suggestion
print(superset(sys.argv[1]))
```

commands/developer-utils/genpasswd.py
```diff
@@ -0,0 +1,80 @@
...
+# Optional parameters:
+# @raycast.icon images/genpasswd.png
+# @raycast.iconDark images/genpasswd-iconDark.png
+
+# Documentation:
+# @raycast.description Generates an iOS style password
+# @raycast.author everdrone
+# @raycast.authorURL https://github.com/everdrone
+
+import string
```
unnamedd, line 18: To make use of argument as suggested
```suggestion
import sys
import string
```

unnamedd (changes requested):
Hi @everdrone, check my comments and let us know what you think.

Thanks for your first contribution in our repository.

everdrone:
> Hi @everdrone, check my comments and let us know what you think.
> 
> Thanks for your first contribution in our repository.

Sure! Thanks for the tips!

But the password length is not exactly defined by one argument in my script. I'll need to make some more changes

### Long explanation:
The password length is defined by this equation: `set_length * length + (len(separator) * length - 1)`

The `length` parameter changes the number of sets (the alphanumeric strings between the separators)
The `set_length` defines the actual length of the substrings in between the separators

The default regexp of the password is `([\w\d]{6}-?){3}`, where `6` is  `set_length` and `3` is `length`

fe9lix (approved):

PR #473, opened 2021-06-12
Add Pulse Secure commands

jaklan:
## Description

Commands to handle VPN connections with Pulse Secure. The solution uses Apple Keychain to store passwords, so you just need to specify it once (and the Keychain item is updated when the password is re-entered). It supports default connection / username (to specify in the command file) and partial matching (e.g. explicit "San Francisco" matches to connection "San Francisco", but query "south" would match to the first matching item, e.g. "South San Francisco"; full match has precedence over partial match).

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

Pulse Secure app

## Checklist


commands/apps/pulse-secure/pulse-secure-connect.applescript
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/osascript
+
+# Dependencies:
+# 	1. Pulse Secure: https://www.pulsesecure.net/
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Connect / Reconnect
```
unnamedd, line 8:
```suggestion
# @raycast.title Connect
```

commands/apps/pulse-secure/pulse-secure-disconnect.applescript
```diff
@@ -0,0 +1,91 @@
...
+	openMenu()
+	tell application "System Events" to tell process "PulseTray"
+		tell menu 1 of menu bar item 1 of menu bar 2
+			if name of menu item 3 does not contain "No active connections" then
+				tell (first menu item whose value of attribute "AXMenuItemMarkChar" is " ")
+					click
+					tell menu 1
+						if value of attribute "AXEnabled" of menu item "Disconnect" then
+							click menu item "Disconnect"
+						else if value of attribute "AXEnabled" of menu item "Cancel" then
```
unnamedd, line 50: Looks to me that Disconnect and Cancel are two different things in Pulse Secure. Am I right? 
If yes, then, will be better if you split this script command in two: one for disconnect and one for cancel the connection, even though you will have a bit of duplicated code.

unnamedd (changes requested):
Hi @jaklan, thanks for your hard use of Apple Script. Really nice code to read!

I've made some suggestion in the code, but also, I'd like to bring another one here: will be amazing if we start to follow a pattern already implemented by some other users submitting a quick and nice README to this folder. Check it out here: https://github.com/raycast/script-commands/tree/master/commands/apps/things

jaklan:
> Hi @jaklan, thanks for your hard use of Apple Script. Really nice code to read!
> 
> I've made some suggestion in the code, but also, I'd like to bring another one here: will be amazing if we start to follow a pattern already implemented by some other users submitting a quick and nice README to this folder. Check it out here: https://github.com/raycast/script-commands/tree/master/commands/apps/things

Yeah, definietly useful for some some complicated scripts (however this particular example don't provide too much additional value over "Description" fields tbh), but I also find it potentially confusing - should I read descriptions [here](https://github.com/raycast/script-commands/tree/master/commands) or look for the README.md file? 

It would be nice if you actually provide automatically generated links to the README.md files under the headings, e.g. here under "Slack": https://github.com/raycast/script-commands/tree/master/commands#slack
so people can still browse the commands in the above list and they can switch to README.md only if it's actually provided, which is clearly marked this way.

commands/apps/pulse-secure/pulse-secure-connect.applescript
jaklan: @unnamedd I would keep it this way, because it indicates what happens when you connect to the already active connection (we have two potential behaviours: reconnect or just ignore).

commands/apps/pulse-secure/pulse-secure-disconnect.applescript
jaklan: @unnamedd yes, they are - "Disconnect" works for the already activated connection, "Cancel" for the connection that is being activated. The reason I put them together is to make the script faster and don't iterate the list of connections (aka menu items) twice - when we run "Connect" we just don't bother about the state of the previous connection, we find the one which is active in any way, we "drop it" and connect to the target connection.

unnamedd:
I guess it's not exactly a matter of confusion, but a nice way to read a bit about what each script command can do without the need to jump into each file to understand what they do. In my opinion, if you copy the script command's description into the readme, would be very nice. Of course, this is not mandatory, but when it was introduced I felt like it was very promising in terms of documentation to everyone.

> It would be nice if you actually provide automatically generated links to the README.md files under the headings, e.g. here under "Slack": https://github.com/raycast/script-commands/tree/master/commands#slack
so people can still browse the commands in the above list and they can switch to README.md only if it's actually provided, which is clearly marked this way.

@jaklan this is an amazing suggestion and for sure I will implement it in the Toolkit! Thanks for the suggestion!

commands/apps/pulse-secure/pulse-secure-disconnect.applescript
unnamedd: As I understood, for you as user, make sense to have them together in the same Script Command, right?

commands/apps/pulse-secure/pulse-secure-connect.applescript
unnamedd: No problem for me, as I don't understand much of what Pulse Secure does, it's me being picky than anything else. :)

commands/apps/pulse-secure/pulse-secure-disconnect.applescript
jaklan: From my 3-year experience with Pulse Secure - imho yes, the connection sometimes takes a bit longer to activate properly and then, when you just want to drop the current connection, you don't want to look at tray just to verify the connection status and wonder if you should run "disconnect" or "cancel" command. 

But of course if anyone actually raises that as an issue one day - we can change it then ;)

jaklan:
@unnamedd README.md added

commands/apps/pulse-secure/pulse-secure-disconnect.applescript
unnamedd: Great! Then, I have no more to add here! 😄

jaklan:
@unnamedd can we merge it then?

unnamedd (approved):
@fe9lix time to merge this one, it's approved to me. ✅

PR #479, opened 2021-06-17
Add character count command

ubuntudroid:
## Description

Counts the characters of either the clipboard or the passed argument.

## Type of change

- [x] New script command

## Screenshot

CleanShot 2021-06-17 at 16 41 00

Needs python 3.x and the [pasteboard library](https://github.com/tobywf/pasteboard).

## Checklist

### DIFF
--- a/commands/developer-utils/count-characters.py
+++ b/commands/developer-utils/count-characters.py
@@ -11,7 +11,7 @@
 # Optional parameters:
 # @raycast.icon 🤖
 # @raycast.argument1 { "type": "text", "placeholder": "Text", "optional": true }
-# @raycast.packageName Developer Utils
+# @raycast.packageName Developer Utilities
 
 # Documentation:
 # @raycast.description Counts the characters of either the clipboard or the passed argument
--- a/commands/developer-utils/count-characters.py
+++ b/commands/developer-utils/count-characters.py
@@ -5,7 +5,7 @@
 
 # Required parameters:
 # @raycast.schemaVersion 1
-# @raycast.title Count characters
+# @raycast.title Count Characters
 # @raycast.mode compact
 
 # Optional parameters:
--- a/commands/developer-utils/count-characters.py
+++ b/commands/developer-utils/count-characters.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python3
+
+# Dependency: This script requires the python `pasteboard` library: https://github.com/tobywf/pasteboard
+# Install via pip: `pip install pasteboard`
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Count characters
+# @raycast.mode compact
+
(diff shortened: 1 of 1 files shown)


ubuntudroid:
Here is a gif that better shows the output of the script (the one in the PR description was trimmed to aggressively and I can't change it anymore:

CleanShot 2021-06-17 at 16 53 11

commands/developer-utils/count-characters.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
+
+# Dependency: This script requires the python `pasteboard` library: https://github.com/tobywf/pasteboard
+# Install via pip: `pip install pasteboard`
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Count characters
```
unnamedd, line 8:
```suggestion
# @raycast.title Count Characters
```

commands/developer-utils/count-characters.py
```diff
@@ -0,0 +1,32 @@
...
+# @raycast.mode compact
+
+# Optional parameters:
+# @raycast.icon 🤖
+# @raycast.argument1 { "type": "text", "placeholder": "Text", "optional": true }
+# @raycast.packageName Developer Utils
```
unnamedd, line 14:
```suggestion
# @raycast.packageName Developer Utilities
```

unnamedd (changes requested):
Hey @ubuntudroid, thanks for submit your Script Command and welcome aboard. 🎉 

Just two small things to change to keep the conformance with the guideline.

unnamedd (approved):
Looks good to me! Thanks for the changes @ubuntudroid!

fe9lix (approved):

jaklan:
@ubuntudroid @unnamedd Dependency on a globally installed pip package is a **very bad** practice. Especially there's absolutely no need to use it, you can just go with plain Bash + Python script with no dependencies, exactly like here:
https://github.com/raycast/script-commands/blob/master/commands/conversions/strikethrough.sh

unnamedd:
@jaklan I didn't say anything about a global pip install, but the point here is the use of `pasteboard` which is clearly not part of the Python framework. Am I wrong?

es183923:
something like the following could replace the script, and it doesn't have any dependencies.
```bash
#!/bin/bash
str=$(pbpaste)
echo ${#str}
```

unnamedd:
> something like the following could replace the script, and it doesn't have any dependencies.
> 
> ```shell
> #!/bin/bash
> str=$(pbpaste)
> echo ${#str}
> ```

For sure, then I’d like to ask one of you to bring this benefit to our community bringing this improvement updating the Script Command! cc @jaklan @es183923

es183923:
Sure!

PR #486, opened 2021-06-21
Count characters command (improves #479)

es183923:
## Description

This PR improves #479. It counts the number of characters of either the clipboard or the given argument without having any dependencies.

## Type of change

no dependencies nor requirements

## Checklist


dehesa:
Good stuff @es183923

jaklan:
@es183923 we can make it even smoother ;)

```bash
if [ -n "$1" ]
then
    str=$1
else
    str=$(pbpaste)
fi
echo "${#str} chars"

es183923:
@jaklan that works too, but I think it's good enough as it is.

PR #491, opened 2021-06-26
Bitwarden: Actually print an error message if no result was found

MarcelBochtler:
## Description

Notify the user if the searched password was not found.

## Type of change

- [ ] New script command
- [x] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Checklist

### DIFF
--- a/commands/password-managers/bitwarden/copy-first-matching-password.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-password.sh
@@ -22,7 +22,7 @@
 # @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
 
 notFound() {
-  echo "The query '$1' did not return a password."
+  echo "The query '${BASH_ARGV[0]}' did not return a password."
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


commands/password-managers/bitwarden/copy-first-matching-password.sh
```diff
@@ -22,7 +22,7 @@
 # @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
 
 notFound() {
-  echo "The query '$1' did not return a password."
+  echo "The query '${BASH_ARGV[0]}' did not return a password."
```
PSalant726, line 25: What does this change accomplish?

PSalant726 (changes requested):
One question, but thanks for fixing the bad ref!

commands/password-managers/bitwarden/copy-first-matching-password.sh
MarcelBochtler: `$1` in the function is the first parameter passed to the _function_ and not the first parameter passed to the _script_.
`${BASH_ARGV[0]}` is the first parameter of the _script_ itself, which is the intended value here.

As an alternative we could call `notFound` with the parameter `$1` => `name=$(echo $item | jq --exit-status ".name") || notFound $1`
What do you prefer?

PSalant726: Ah, of course. That makes sense. I prefer what you have here.

PSalant726 (approved):
Thanks for the changes!

PR #497, opened 2021-07-06
Fix ip scripts with pbcopy

Seypopi:
## Description

Fix ip script because it was not copying to clipboard

## Type of change

## Checklist


mattisssa (approved):
Hah, how it could pass the review previous time 😁
Thanks for the fix!

es183923:
@mattisssa Now, the code updates your clipboard once an hour. The code needs to be updated so that either it doesn't copy to your clipboard, or not automatically update.

mattisssa:
> @mattisssa Now, the code updates your clipboard once an hour. This is something that should **not** happen. The code needs to be updated so that either it doesn't copy to your clipboard, or not automatically update.

Yeah, thanks for heads up on it 👍

PR #498, opened 2021-07-06
Add move desktop items to trash command

Seypopi:
## Description

It moves all items (folders and files) to the trash.
User confirmation is needed before execution.

## Type of change

- [X] New script command

None

## Checklist


commands/system/move-desktop-to-trash.applescript
```diff
@@ -0,0 +1,22 @@
...
+
+# Documentation:
+# @raycast.author Seypopi
+# @raycast.authorURL https://github.com/Seypopi
+# @raycast.description Empty the desktop.
+
+tell application "Finder"
+    set allDesktop to every item of folder (path to Desktop folder as text)
+    move allDesktop to trash
+    log ""
```
mattisssa, line 21: Why do we need it at the end? @Seypopi

mattisssa (commented):
Was afraid to test it 😄 
Looks fine, but I wonder about the `log` you have in the code

commands/system/move-desktop-to-trash.applescript
Seypopi: Ahah I understand it, but don't worries every think will be waiting in the trash 🙃

The `log ""` is used here to avoid getting the last line in overlaying HUD toast

> In silent mode the last line (if exists) will be shown in overlaying HUD toast after Raycast window is closed.

PR #502, opened 2021-07-08
TOTP support for Bitwarden

GraysonCAdams:
## Description

Adds TOTP clipboard support for the Bitwarden script command. Uses a Python3 library and virtualenv to avoid affecting the macOS packages.

## Type of change

- [x] New script command

## Screenshot

image

No dependencies. macOS comes with Python 3. Installs virtualenv automatically, sets up the virtual environment, and then installs minitotp.

## Checklist

### DIFF
--- a/commands/password-managers/bitwarden/copy-first-matching-totp.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-totp.sh
@@ -3,11 +3,8 @@
 # Dependencies:
 #   1. The Bitwarden CLI: https://bitwarden.com/help/article/cli/
 #   2. The `jq` utility: https://stedolan.github.io/jq/
-#   3. Python3: https://programwithus.com/learn/python/install-python3-mac
-#   4. Python3 virtualenv: https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/
 #
 # Install via homebrew: `brew install bitwarden-cli jq`
-# Follow instructions on steps 3-4
 
 # Required parameters:
 # @raycast.schemaVersion 1
@@ -23,24 +20,6 @@
 # @raycast.author Phil Salant
 # @raycast.authorURL https://github.com/PSalant726
 # @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
-
-# Activate Python environment
-if ! command -v virtualenv &> /dev/null
-then
-  echo "virtualenv command is required";
-  exit 1;
-fi
-
-if [ ! -d virtualenv ]; then
-  virtualenv venv
-fi
-
-source venv/bin/activate
-
-if ! command -v mintotp &> /dev/null
-then
-    pip3 install mintotp # installs in virtualenv
-fi
 
 notFound() {
   echo "The query '${BASH_ARGV[0]}' did not return a TOTP."
@@ -71,12 +50,8 @@ if [ $unlocked_status -ne 0 ]; then
   exit 1
 fi
 
-item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { name: .name, totp: .login.totp }")
-name=$(echo $item | jq --exit-status ".name") || notFound
-totp=$(echo $item | jq --raw-output --exit-status ".totp") || notFound
-totp=$(echo $totp | sed -E "s/^(.)+\?secret=//g" | sed -E "s/&issuer=(.)+//g") || notFound
-
-echo -n $(mintotp <<< $totp) | pbcopy
+totp=$(bw get totp "$1" $session)
+echo -n $totp | pbcopy
 unset totp
 echo "Copied the TOTP for '$name' to the clipboard."
 exit 0
--- a/commands/password-managers/bitwarden/copy-first-matching-totp.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-totp.sh
@@ -50,7 +50,7 @@ if [ $unlocked_status -ne 0 ]; then
   exit 1
 fi
 
-totp=$(bw get totp "$1" $session)
+totp=$(bw get totp "$1" $session) || notFound
 echo -n $totp | pbcopy
 unset totp
 echo "Copied the TOTP for '$1' to the clipboard."
--- a/commands/password-managers/bitwarden/copy-first-matching-totp.sh
+++ b/commands/password-managers/bitwarden/copy-first-matching-totp.sh
@@ -3,8 +3,11 @@
 # Dependencies:
 #   1. The Bitwarden CLI: https://bitwarden.com/help/article/cli/
 #   2. The `jq` utility: https://stedolan.github.io/jq/
+#   3. Python3: https://programwithus.com/learn/python/install-python3-mac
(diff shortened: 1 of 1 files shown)


jaklan:
- There's no need for `virtualenv` as a dependency - you have built-in `venv`:
https://docs.python.org/3/library/venv.html
- You create the virtual environment in the directory where the script is located with the very common name `venv`. It can lead to unexpected errors and conflicts with other commands.
- In general I recommend to use `pipx` for installation of such dependencies, either `pipx install` or `pipx run`, instead of doing that manually.

But the most important note - is that dependency actually needed? I don't have the premium account, so I can't test it, but it seems that `bw get totp` already does the job:
https://bitwarden.com/help/article/cli/#get

jaklan:
Ah okay, I see what you did here - you enter TOTP secret key in Bitwarden, but use `mintotp` to generate TOTP value, so you don't have to have the premium account. Nice workaround 😄 Btw it would be valuable to provide some option for actual Premium users to avoid the dependency and just run proper Bitwarden command then.

GraysonCAdams:
> Ah okay, I see what you did here - you enter TOTP secret key in Bitwarden, but use `mintotp` to generate TOTP value, so you don't have to have the premium account. Nice workaround 😄 Btw it would be valuable to provide some option for actual Premium users to avoid the dependency and just run proper Bitwarden command then.

Updated.

That was unintentional. I was unaware that the BW CLI has TOTP support. I think it makes most sense to use that, and people can adapt the code if necessary using a similar method as I did such as `mintotp`.

Fair points about not needing `virtualenv` plus using `pipx`.

I am not sure that using `source venv/bin/activate` or generating a folder with that name would really cause issues with this script... however `python3 -m venv` would make the most sense, or as you suggested, `pipx`. Regardless, the BW CLI is the real answer, I feel.

jaklan:
@GraysonCAdams the problem is with the location - people can put Raycast scripts pretty everywhere, and location of the script is used as a working directory for it. Automatically creating a Python venv with definitely-not-unique name like `venv` in such a location can cause unexpected issues if e.g. there was already existing `venv` directory there.

jaklan:
@GraysonCAdams one more note - if `bw list items --search` provide only TOTP key, not TOTP value, I would use it anyway just to get the full name of the item which was used, as in the Phil's script:
https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/copy-first-matching-password.sh
Currently, your output shows the initial query, so you don't know what was the final match.

mattisssa (changes requested):
Thanks for the PR and looks like Jakub already reviewed the extension and code you've written, thanks to a lot mate @jaklan! 
I would recommend changing at least the issue with `venv` directory

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,57 @@
...
+if [ $unlocked_status -ne 0 ]; then
+  echo "Vault is locked!"
+  exit 1
+fi
+
+totp=$(bw get totp "$1" $session) || notFound
+echo -n $totp | pbcopy
+unset totp
+echo "Copied the TOTP for '$1' to the clipboard."
+exit 0
```
PSalant726, line 57: I question the user experience here. The `bw get` command will return a single result, but it's unclear to me how that result is determined. If a user has multiple tokens that would each match the same search term, there is no way of knowing which token was copied _and_ the output here would just be whatever term was initially searched.

For example, if a user has multiple AWS logins, each with their own token, then searching for `AWS` as the value of `$1` could theoretically return any of 3 different tokens. Still, the output would just read "Copied the TOTP for AWS to the clipboard."

This is precisely why I chose the `bw list items --search` command when implementing [`copy-first-matching-password.sh`](https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/copy-first-matching-password.sh#L53). In my script, a user can search for any term they desire, but the output reflects the name of the search _result_ who's password was copied to the clipboard. Unless we can find a way to achieve the same effect here, this script should not be merged.

Have you tested the following?

```suggestion
item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { name: .name, totp: .login.totp }")
name=$(echo $item | jq --exit-status ".name") || notFound
totp=$(echo $item | jq --raw-output --exit-status ".totp") || notFound

echo -n $totp | pbcopy
unset totp
echo "Copied the one-time password for '$name' to the clipboard."
exit 0
```

I don't have TOTP tokens saved in my personal Bitwarden vault, but from looking at the JSON output of `bw list items`, I believe it should work.

PSalant726 (changes requested):
Thank you for the PR! Unfortunately there is one major issue, which I've outlined in a comment below.

Separately, can you please add this script to the package's [`README.md` file](https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/README.md) as a part of these changes?

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,57 @@
...
+# @raycast.title Copy First Matching TOTP
+# @raycast.mode silent
+#
+# Optional parameters:
+# @raycast.packageName Bitwarden
+# @raycast.icon images/bitwarden.png
+# @raycast.argument1 { "type": "text", "placeholder": "Query" }
+#
+# Documentation
+# @raycast.author Phil Salant
```
PSalant726, line 20: You can duplicate this line and add your own name. There is support for multiple authors!

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,57 @@
...
+# @raycast.authorURL https://github.com/PSalant726
+# @raycast.description Search all items in a Bitwarden vault, and copy the password of the first search result to the clipboard.
```
PSalant726, line 22:
```suggestion
# @raycast.description Search all items in a Bitwarden vault, and copy the time-based one-time password (TOTP) of the first search result to the clipboard.
```

PSalant726 (changes requested):

commands/password-managers/bitwarden/copy-first-matching-totp.sh
jaklan: @PSalant726 please read my comment above.
```
name=$(bw get item "$1" $session 2> /dev/null | jq --exit-status ".name") || notFound
totp=$(bw get totp "$name" $session) || notFound
```
also should do the job and I would also recommend to use `bw get item` in your scripts.

PSalant726: Ha, thanks for the edit. I came back to say something similar. If you have more than one vault item with `$name`, then you can't be certain which TOTP will be returned.

int3rrupt: If you have more than one item with `$name` then I believe it throws an error.

int3rrupt:
I was going to PR my TOTP implementation but then I noticed that this PR was out there. My approach was to start with the _Copy First Matching Password_ script as a base. I created a draft PR in case anyone wants to take a look: #508.

PR #508, opened 2021-07-13
Add Bitwarden TOTP command

int3rrupt:
## Description

This adds a new script that allows Bitwarden users to copy TOTP codes to their clipboards.

## Type of change

- [x] New script command
- [x] Documentation update

## Screenshot

copy-first-matching-totp

All commands in the Bitwarden package already require Bitwarden CLI, including this one. This script, like some of the existing Bitwarden scripts, also requires the jq utility. Both are installed via homebrew.

## Checklist


commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,66 @@
...
+# @raycast.title Copy First Matching TOTP
+# @raycast.mode silent
+#
+# Optional parameters:
+# @raycast.packageName Bitwarden
+# @raycast.icon images/bitwarden.png
+# @raycast.argument1 { "type": "text", "placeholder": "Query" }
+#
+# Documentation
+# @raycast.author Adrian Reyes
```
PSalant726, line 20: There's support for multiple authors by duplicating this line. Since this is mostly code from the _Copy First Matching Password_ script, I would appreciate a nod 😄 . This isn't a blocker for me.
```suggestion
# @raycast.author Adrian Reyes
# @raycast.author Phil Salant
```

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,66 @@
...
+
+if [ $unlocked_status -ne 0 ]; then
+  echo "Vault is locked!"
+  exit 1
+fi
+
+item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { id: .id, name: .name }")
+name=$(echo $item | jq --exit-status ".name") || itemNotFound
+id=$(echo $item | jq --exit-status ".id") || itemNotFound
+totp=$(eval "bw get totp $id $session" 2> /dev/null) || itemNoTotp "$name"
```
PSalant726, line 61: There's absolutely nothing functionally wrong with this, as far as I can tell, but the `bw list items` command will return the TOTP in it's result. If we copy it from there instead of the `id`, then we can make one less call to the CLI.
```suggestion
item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { name: .name, totp: .login.totp }")
name=$(echo $item | jq --exit-status ".name") || notFound
totp=$(echo $item | jq --raw-output --exit-status ".totp") || notFound
```

PSalant726 (changes requested):
Thanks for the PR! I have just two very minor suggestions for code changes.

Separately, can you please add this script to the package's [`README.md` file](https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/README.md) as a part of these changes? It probably belongs right after _Copy First Matching Password_.

commands/password-managers/bitwarden/copy-first-matching-totp.sh
int3rrupt: The actual TOTP isn't provided in the JSON which is why we need `bw get totp`. What you get in the JSON is the TOTP key/secret that is used to generate the TOTP. I also don't see any option to include it in the JSON output. Have you seen it in the JSON? Am I missing something?

PSalant726: Ah, yes that makes more sense than assuming the `bw list items` JSON output would change every 30 seconds 😂 . I haven't seen any other TOTP-related fields in the JSON, but it doesn't matter anyway because what you have should work just fine.

PSalant726 (changes requested):
I don't have TOTP tokens in my personal Bitwarden vault, so if you've been able to test your code as-is, then it works for me (although I'm not sure you need [the `eval`](https://github.com/raycast/script-commands/pull/508/files#diff-ec14dee94b8d0225cbb35f7716146e488b29bcf4a5553b450f36c44354bea3f5R62)).

Separately, can you please add this script to the package's [`README.md` file](https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/README.md) as a part of these changes? It probably belongs right after _Copy First Matching Password_.

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,67 @@
...
+# @raycast.author Phil Salant
+# @raycast.authorURL https://github.com/int3rrupt
+# @raycast.description Search all items in a Bitwarden vault, and copy the TOTP of the first search result to the clipboard.
+
+itemNotFound() {
+  echo "The query '${BASH_ARGV[0]}' did not return any items."
+  exit 1
+}
+
+itemNoTotp() {
```
jaklan, line 30: To keep it consistent, I would rename it to "totpNotFound":

```suggestion
totpNotFound() {
```

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,67 @@
...
+  echo "Item '$1' does not have a TOTP configured."
```
jaklan, line 31:
```suggestion
  echo "The item '$1' does not have a TOTP configured."
```

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,67 @@
...
+
+if [ $unlocked_status -ne 0 ]; then
+  echo "Vault is locked!"
+  exit 1
+fi
+
+item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { id: .id, name: .name }")
+name=$(echo $item | jq --exit-status ".name") || itemNotFound
+id=$(echo $item | jq --exit-status ".id") || itemNotFound
+totp=$(eval "bw get totp $id $session" 2> /dev/null) || itemNoTotp "$name"
```
jaklan, line 62: As @PSalant726 mentioned, is there any reason to use eval here?

```suggestion
totp=$(bw get totp $id $session 2> /dev/null || totpNotFound "$name")
```

jaklan (changes requested):

int3rrupt:
> I don't have TOTP tokens in my personal Bitwarden vault, so if you've been able to test your code as-is, then it works for me (although I'm not sure you need [the `eval`](https://github.com/raycast/script-commands/pull/508/files#diff-ec14dee94b8d0225cbb35f7716146e488b29bcf4a5553b450f36c44354bea3f5R62)).
> 
> Separately, can you please add this script to the package's [`README.md` file](https://github.com/raycast/script-commands/blob/master/commands/password-managers/bitwarden/README.md) as a part of these changes? It probably belongs right after _Copy First Matching Password_.

Originally I didn't have `eval` but I wasn't getting a token while running it through Raycast, however, executing the actual command in the terminal works. I'm not very familiar with bash scripting so it might be a simple fix.

This is what I had before:
```bash
totp=$(bw get totp $id $session 2> /dev/null) || itemNoTotp "$name"
```
I think the issue is due to the way I am referencing the variables. Once I wrap the command in a string and throw an `eval` in front, it works as expected. I too would prefer not having that `eval` in there, any suggestions?

commands/password-managers/bitwarden/copy-first-matching-totp.sh
int3rrupt: See my response in the comment below. If we can get rid of it that would be great.

PSalant726:
Since I can't test it locally, it's unclear what `bw get totp` command returns, but the fact that wrapping the call in a string seems to work implies that it's just returning the raw TOTP code. You can instead try wrapping the entire `$()` in a string, without `eval`, like:
```sh
totp="$(bw get totp $id $session 2> /dev/null) || itemNoTotp '$name')"
```

int3rrupt:
> Since I can't test it locally, it's unclear what `bw get totp` command returns, but the fact that wrapping the call in a string seems to work implies that it's just returning the raw TOTP code. You can instead try wrapping the entire `$()` in a string, without `eval`, like:
> 
> ```shell
> totp="$(bw get totp $id $session 2> /dev/null) || itemNoTotp '$name')"
> ```

This did not work. I also removed the extra parenthesis:
```bash
totp="$(bw get totp $id $session 2> /dev/null || itemNoTotp '$name')"
```
This did not work either. Other ideas?

PSalant726:
Ah, sorry the typo was the last `)`, not the first one. So it should have been:
```sh
totp="$(bw get totp $id $session 2> /dev/null)" || itemNoTotp '$name'
```
This way, you'll place the return value of `bw get totp` into a string `||` the string returned by `itemNoTotp`. This is all just a guess. I'm not a bash expert either 😅

int3rrupt:
> Ah, sorry the typo was the last `)`, not the first one. So it should have been:
> 
> ```shell
> totp="$(bw get totp $id $session 2> /dev/null)" || itemNoTotp '$name'
> ```
> 
> This way, you'll place the return value of `bw get totp` into a string `||` the string returned by `itemNoTotp`. This is all just a guess. I'm not a bash expert either 😅

I believe this is close to what I was originally trying when I reached out on Slack. I did just try this though to confirm. It appears to work, because I don't the the error message, however, I also don't get a value in my clipboard.

PSalant726:
Ok, I think it's because `""` isn't treated as "falsey" in Bash like it would be in JS, so you're just getting `""` on your clipboard. This pattern probably wont work for `bw get totp` like it does for `jq --exit-status`. You may need to try refactoring it to check the exit code of `bw get totp` after calling it, and printing the error message unless the exit code is 0. I've implemented something similar on L54 for `bw unlock `.

int3rrupt:
Ok, I'll check it out. Thanks.

int3rrupt:
`eval` has been removed. I just needed to add `-r` to the `jq` command in order to remove the quotes from the ID. Also added check for exit code.

commands/password-managers/bitwarden/README.md
```diff
@@ -25,7 +26,7 @@ A package of script commands to interact with [Bitwarden Vaults](https://bitward
 
 ## Dependencies
 
-All authentication and vault-related commands in this package require the [Bitwarden CLI](https://bitwarden.com/help/article/cli/) v1.14.0 or later. Bitwarden Send-related commands require v1.15.1 or later. The _Bitwarden Status_, _Search Vault Items_, _Copy First Matching Password_, and Bitwarden Send-related commands (except _Delete a Send_) also require the [`jq` utility](https://stedolan.github.io/jq/).
+All authentication and vault-related commands in this package require the [Bitwarden CLI](https://bitwarden.com/help/article/cli/) v1.14.0 or later. Bitwarden Send-related commands require v1.15.1 or later. The _Bitwarden Status_, _Search Vault Items_, _Copy First Matching Password_,, _Copy First Matching TOTP_ and Bitwarden Send-related commands (except _Delete a Send_) also require the [`jq` utility](https://stedolan.github.io/jq/).
```
PSalant726, line 29: Just a misplaced `,` character here.
```suggestion
All authentication and vault-related commands in this package require the [Bitwarden CLI](https://bitwarden.com/help/article/cli/) v1.14.0 or later. Bitwarden Send-related commands require v1.15.1 or later. The _Bitwarden Status_, _Search Vault Items_, _Copy First Matching Password_, _Copy First Matching TOTP_, and Bitwarden Send-related commands (except _Delete a Send_) also require the [`jq` utility](https://stedolan.github.io/jq/).
```

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,73 @@
...
+unlocked_status=$?
+
+if [ $unlocked_status -ne 0 ]; then
+  echo "Vault is locked!"
+  exit 1
+fi
+
+item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { id: .id, name: .name }")
+name=$(echo $item | jq --exit-status ".name") || itemNotFound
+id=$(echo $item | jq -r --exit-status ".id") || itemNotFound
```
PSalant726, line 61: Please use the long-form of `-r`. It's a hard-coded command, so it improves the maintainability if other contributors can get an idea of what it's doing without having to look at documentation.
```suggestion
id=$(echo $item | jq --raw-output --exit-status ".id") || itemNotFound
```

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,73 @@
...
+totp=$(bw get totp $id $session 2> /dev/null)
+status=$?
```
PSalant726, line 63:
```suggestion
totp_status=$?
```

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,73 @@
...
+
+if [ $status -eq 0 ]; then
+  echo -n $totp | pbcopy
+  unset totp
+  echo "Copied the TOTP for '$name' to the clipboard."
+  exit 0
+else
+  totpNotFound "$name"
+  exit 1
+fi
```
PSalant726, line 73: We can flatten this a bit.
```suggestion
if [ $totp_status -ne 0 ]; then
  totpNotFound "$name"
  exit 1
fi

echo -n $totp | pbcopy
unset totp
echo "Copied the TOTP for '$name' to the clipboard."
exit 0
```

PSalant726 (changes requested):
Glad you got got this working! Thanks for sticking with it, and thanks for updating the `README.md` as well.

My only remaining comments are just some small stylistic improvements.

int3rrupt:
Suggestions applied.

PSalant726 (approved):
LGTM! Thanks again!

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,73 @@
...
+
+item=$(bw list items --search "$1" $session 2> /dev/null | jq ".[0] | { id: .id, name: .name }")
+name=$(echo $item | jq --exit-status ".name") || itemNotFound
+id=$(echo $item | jq --raw-output --exit-status ".id") || itemNotFound
+totp=$(bw get totp $id $session 2> /dev/null)
+totp_status=$?
+
+if [ $totp_status -ne 0 ]; then
+  totpNotFound "$name"
+  exit 1
```
jaklan, line 67:
```suggestion
```
It's redundant here, because you are already exiting in `totpNotFound` function

jaklan: In general, you can replace lines 63-68 with one-liner:
```
test $? -eq 0 || totpNotFound "$name"
```

jaklan (changes requested):

commands/password-managers/bitwarden/copy-first-matching-totp.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
```
jaklan, line 68: In general, you can replace lines 63-68 with one-liner:
```
test $? -eq 0 || totpNotFound "$name"
```

jaklan (changes requested):

int3rrupt:
Suggestion applied and tested.

jaklan (approved):
Imho it's ready to merge, @mattisssa or anyone?

mattisssa:
@jaklan nice 👍

PR #515, opened 2021-07-17
Add Script Command: Trim Git Commit Hash

crstauf:
## Description

Accept parameter of full git commit hash and trim down to seven characters.

## Type of change

- [x] New script command

## Screenshot

Screen Shot 2021-07-16 at 10 05 32 PM

None

## Checklist

### DIFF
--- a/commands/developer-utils/trim-git-commit-hash.sh
+++ b/commands/developer-utils/trim-git-commit-hash.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# @raycast.title Trim Git Commit Hash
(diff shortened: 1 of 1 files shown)


grzegorzkrukowski (changes requested):
Hey - thanks for your contribution.

I think it will be worth adding information about the fact that value was copied to the clipboard. 
You can either do:
`echo "Copied ${str:0:7}"`
or just
`echo "Copied to clipboard"`

at the very end of the script

grzegorzkrukowski (approved):

PR #516, opened 2021-07-18
Add Obsidian search vault and create note templates

HotThoughts:
## Description

Add two templates for the Obsidian app
- search in a given vault, and
- create a new note in the given vault

## Type of change

- [X] New script command

## Screenshot

When creating a new note, `Name` of the note is required, while `Content` is optional.
image

This will start a search query in the given vault.
image

Requires Obsidian `v0.8.15+` as this is the version that [URL scheme was first introduced](https://forum.obsidian.md/t/obsidian-release-v0-8-15/5697). 
You can Install Obsidian via: 
1. the offical website: https://obsidian.md
2. use HomeBrew: `brew install --cask obsidian`

## Checklist


grzegorzkrukowski:
Thanks for your first contribution! 

In general, it all looks good, but I would like to suggest adding a check for `$VAULT_NAME` being empty in each of the scripts:

```
VAULT_NAME=""

if [ -z "$VAULT_NAME" ]
then
  echo "Configure VAULT_NAME"
  exit 1
fi
```

In case someone misses the fact he should do it directly in the script file - it's gonna give a bit nicer error handling than the current solution that is error shown directly by Obsidian

grzegorzkrukowski (changes requested):
Thanks for your first contribution! 

In general, it all looks good, but I would like to suggest adding a check for `$VAULT_NAME` being empty in each of the scripts:

```
VAULT_NAME=""

if [ -z "$VAULT_NAME" ]
then
  echo "Configure VAULT_NAME"
  exit 1
fi
```

In case someone misses the fact he should do it directly in the script file - it's gonna give a bit nicer error handling than the current solution that is error shown directly by Obsidian

HotThoughts:
@grzegorzkrukowski Woohoo my first contribution  :tada: 

I added the suggested error handling. 😄

grzegorzkrukowski (approved):

PR #517, opened 2021-07-20
Add stopwatch script command

pomdtr:
## Description

Add a stopwatch mechanism to raycast. You can start, stop and see progress of an active stopwatch

- [x] New script command

## Screenshot

Not much to see.

## Checklist


commands/productivity/stopwatch/start.bash
```diff
@@ -0,0 +1,23 @@
...
+# @raycast.icon ⏱
+# @raycast.packageName time
+
+# Documentation:
+# @raycast.description Start a stopwatch
+# @raycast.author Achille Lacoin
+# @raycast.authorURL https://github.com/pomdtr
+
+if [ -f  timestamp_start.txt ]; then
+    echo "A stopwatch already exists!" > /dev/stderr
```
thomaspaulmann, line 18: Good one!

commands/productivity/stopwatch/stop.bash
```diff
@@ -0,0 +1,24 @@
...
+# @raycast.packageName time
+
+# Documentation:
+# @raycast.description Stop active stopwatch, copy total time
+# @raycast.author Achille Lacoin
+# @raycast.authorURL https://github.com/pomdtr
+
+set -e
+
+PROGRESS=$(./progress.sh)
```
grzegorzkrukowski, line 19: There is no `progress.sh` file included in this PR - should this be `.bash` instead here?

grzegorzkrukowski (changes requested):

grzegorzkrukowski:
I would also suggest adding `stopwatch-` prefix to all filenames - some users will throw all scripts into one folder.

commands/productivity/stopwatch/stop.bash
pomdtr: Oups. I'll correct it

pomdtr: Done in da945f7

es183923:
I think putting the file in `/tmp/stopwatch/stopwatch.txt` or somewhere similar would be a better location for the stopwatch temporary file.

pomdtr:
I agree, I moved it to the system TMPDIR in 82d69ea.

Thanks for the feedback!

jaklan:
@pomdtr it's not as issue actually, but `.bash` suffix is pretty rare (even if it's explicitly `bash`, not `sh`, script), I would keep with the standard `.sh` as all other bash scripts in the repository.

grzegorzkrukowski:
@pomdtr why we ended up having both .sh and .bash in this PR ?

grzegorzkrukowski (changes requested):
@pomdtr why we ended up having both .sh and .bash in this PR ? Let's remove duplicated files and leave .sh only

grzegorzkrukowski (approved):
Looks good now - thanks for the changes

PR #518, opened 2021-07-20
Change "colour" to "color"

es183923:
## Description

Swift uses "color", so I felt that using modifying it to be consistent would be better. Also, I renamed the title so that typing "sample color" or "sample colour" would work.

## Type of change

## Checklist

### DIFF
--- a/commands/system/sample-color.swift
+++ b/commands/system/sample-color.swift
@@ -2,7 +2,7 @@
 
 // Required parameters:
 // @raycast.schemaVersion 1
-// @raycast.title Sample Colour
+// @raycast.title Sample Color (Sample Colour)
 // @raycast.mode silent
 // @raycast.packageName System
 //
@@ -10,7 +10,7 @@
 // @raycast.icon 🎨
 //
 // Documentation:
-// @raycast.description Sample a colour from anywhere on your screen.
+// @raycast.description Sample a color from anywhere on your screen.
 // @raycast.author Jesse Claven
 // @raycast.authorURL https://github.com/jesse-c
 
@@ -30,9 +30,9 @@ extension NSColor {
   }
 }
 
-func copyToPasteboard(_ colour: String) {
+func copyToPasteboard(_ color: String) {
   NSPasteboard.general.clearContents()
-  NSPasteboard.general.writeObjects([colour as NSPasteboardWriting])
+  NSPasteboard.general.writeObjects([color as NSPasteboardWriting])
 }
 
 let sampler = NSColorSampler()
@@ -41,10 +41,10 @@ sampler.show { selectedColor in
   if let selectedColor = selectedColor {
     let hexTuple = selectedColor.hexAlphaString
     copyToPasteboard(hexTuple)
-    print("Sampled colour: \(hexTuple)")
+    print("Sampled color: \(hexTuple)")
     exit(0)
   } else {
-    print("Sampled colour: none")
+    print("Sampled color: none")
     exit(0)
   }
 }
--- a/commands/system/sample-color.swift
+++ b/commands/system/sample-color.swift
@@ -2,7 +2,7 @@
 
 // Required parameters:
 // @raycast.schemaVersion 1
-// @raycast.title Sample Color (Sample Colour)
+// @raycast.title Sample Color
 // @raycast.mode silent
 // @raycast.packageName System
 //


commands/system/sample-color.swift
```diff
@@ -2,15 +2,15 @@
 
 // Required parameters:
 // @raycast.schemaVersion 1
-// @raycast.title Sample Colour
+// @raycast.title Sample Color (Sample Colour)
```
jaklan, line 5:
```suggestion
// @raycast.title Sample Color
```

I wouldn't complicate it and just keep one version, `color`, as also used [here](https://raycast.com/features):
image

jaklan (changes requested):

unnamedd:
To be really honest, I don't agree with the rename, even if NSColor doesn't have the "u", the original author created the script on his way for a reason (I'd assume he is British) and this way make sense to him, rename to "Color" will make American english speaker feel better, but not the British ones. In my opinion It is just a matter of respect for the original name and the author.

Ps: he could easily have created a typealias to have NSColour in his language, but he preferred just to use the name on his way.
```swift
typealias NSColour = NSColor
```

Also, no one need to type: "Sample Colour", if you just type "Sample Colo" Raycast will find it!

#my2cents

jaklan:
@unnamedd tbh I totally disagree with that one. There are too many differences between British English and American English, if we don't introduce any rules here - we will end up with a mess. Predictability & consistency are imho very important factors in such community-driven initiatives. On the official Raycast website American English is used, so that's why I assume it's should be "the default English" , but generally it should be pointed out in the repo docs which one should be used.

unnamedd:
This is the beauty of the democracy, I disagree with the rename just for the rename and you disagree with me and that's completely okay.

To proceed with this PR I'd highly recommend bring to this thread the opinion of the original author. To me, is a matter of respect. I would be very disappointed if someone just get something I worked on and just change without even consider to get in touch to me.

We are a community and we respect each other, therefore let's ping the author before continue the discussion.

jaklan:
@unnamedd tagging original authors when modifying existing scripts should be in general the default behaviour (and imho it should be pointed out in `CONTRIBUTING.md` explicitly), because we can miss some context when providing changes.

But I don't agree that general rules about the language used should be the matter of the original author's taste. I don't feel offended I have to write everything in English, even if I would prefer to do it in Polish / French / Chinese / whatever. There should be simple statement in `CONTRIBUTING.md` like _Please use American English_ or _Please use British English_ and the issue is solved for ever.

unnamedd:
@jaklan, if you are changing the implementation, to improve some weird behaviour, I completely agree with you, but isn't the case here, what the PR is proposing is the rename of title, which is in british english and not in american english. That is my point. If we are proposing this change, we must yes, get in touch to the author to ask his opinion.  We always knew about the difference of "color" and "colour" and even with that in mind, decided to go to "colour".

As a reviewer, I already saw a lot os implementations changes here, being made by someone else which was not the original author and I never brought this point to the discussion, however, as we are pushing for a rename, for a matter of respect, I'd consult the original author. Probably, he would mind the change and this a win-win, but what about if he mind? Are we going to just skip his opinion? As you mentioned and I agree with you is we don't have it described in the `CONTRIBUTING.md`. I know you would prefer to write in your monther tongue language, as I would do either, but we are in an international community and we must respect the english as an universal language which our languages unfortunately are not.

jaklan:
@unnamedd I brought the example of other languages just to show that, when working on publicly available stuff for a broad community, you accept some general rules for "the greater good" and don't mind your individual preferences. Of course now it's a "gray area", so contributors use whatever English they prefer, but that's why I say such a topic should be covered in `CONTRIBUTING.md` to end any unnecessary discussions. And then I find completely nothing wrong to send such PRs to allign scripts to some common standard. If it's missing, as in this case - that's the perfect moment to take a decision ;)

unnamedd:
And I agree with your concerns and "the greater good" and also with your points regarding the english language, however, this is something we are already doing in every ground of Raycast, your mother tongue is different than mine and we are here, discussing in english and we don't care which english we are using here, the only point that I don't agree is with any imposition about something as you mentioned is in a grey area. The variant of the english used throughout Raycast wasn't explicitly defined, therefore, there's not yet any clear definition which one is right or wrong. American english or British english are totally correct and fine to use, however, we have a clear definition of the preference of the author when he implemented this Script Command we are discussing about and I'm sure he won't mind if we decide to standardise the Script Commands using the American english, but first, I really think it very important to check with him before, to show him the respect we would like to have to ourselves.

I do think that the discussion we are having here, would might happen in the future anyway, fortunately we are here, deciding together and in a public way and I'm pretty sure it won't be needed to happen in the future anymore.

Just to finish, I do not think that is wrong to send such PRs standardising some Script Commands to the same language, but not before this discussion here. This must be aligned between everyone, appended a note into the `contributing.md` file and then move on. In my opinion, this is the only way we have to have any standard. Proposing, discussing and then implementing.

jaklan:
Definitely, that's why imho it would be good to provide some conclusion from that discussion to the docs, so it wouldn't be needed in the future - even if the final outcome is just "use whatever English you prefer".

In general I have an impression we are mixing here 2 topics - respect for contributors and repo standards, and unnecessarily opposed them to one another ;) I fully agree with your point, I'm just looking at this issue not from the perspective of changing other people' scripts against their will, but as an excuse to discuss some potential standards and then, after making the final decision, just treat such PRs as a step toward alignment. Ofc it would be better to start such a discussion on Slack before sending a PR, because more people can share their voice, but generally just creating it and discussing common approach on the example of some concrete script shouldn't be treated as anything disrespectul for the author.

Btw I have just noticed you have a "Collaborator", and not "Contributor", badge, so I guess you represent the Raycast team - I was sure I was just discussing with some other contributor and we were still waiting for the feedback from the team ;)

unnamedd:
And we are at the same side in the matter of the standard! I completely agree we should have to start this discussion somewhere in time, I just didn't know it would come so fast! haha

To be honest, as this repository is open to anyone, my opinion is: if we open the discussion here, we would make it even wider to more people bring their thoughts than inside of Slack, which is more closed to people who are there at the moment we are discussing, here all our points will stay for the eternity.

Now, I guess our next step is to wait a bit for a position of the Raycast team, after all, we are just community members, we bring the points, they decide!

To finish, thanks a lot to bring your points, I do think all the perspective brought to the discussion are totally valid.

PitNikola:
Hey folks. Going to duplicate here [my response]((https://raycastcommunity.slack.com/archives/C019B6CGX0T/p1626894664201100?thread_ts=1626862234.195700&cid=C019B6CGX0T)) from the Slack Community:

At the moment, when there is no such guideline in CONTRIBUTING.md, it’s hard to enforce it (and to reason about it). Personally I’m inclining towards enforcing American English language in this particular repo since it’s more common language, especially in the dev world.

We’ll discuss it internally and get back to you with our decision. For time being let's keep this PR open.

grzegorzkrukowski:
Hey guys - we had a long internal discussion about this problem and we decided to stick with American English as a convention for this repository.

Thanks for bringing up the problem and for your input to the discussion.

We added `English style convention` to our [CONTRIBUTING.md](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#english-style-convention) to make it official.

At the same time - I am gonna merge this PR - we will review the rest of the scripts later on and update them to follow it up.

unnamedd:
Excellent! Then we were already good to go! Thanks for the merge @grzegorzkrukowski

PR #531, opened 2021-08-03
added clear cache and reload script

aaronhmiller:
## Description

This script use the option-command-R keyboard shortcut to force refresh a Safari page. It arose out of a need to get a current CI badge status when using Github Action workflows because I noticed that the badge would not reflect the current status unless cache was cleared and the page refreshed. Normally that requires enabling Developer Menu and opt-cmd-E and cmd-R, but opt-cmd-R seems to do both and doesn't require enabling Developer Mode. Having this in Raycast is just one less thing to remember and it's nice to have it stored centrally.

## Type of change

- [X] New script command

## Screenshot

Not much to see here...it just force refreshes the frontmost safari page. I added a comment about this being undocumented and therefore subject to change w/o notice. I also noticed that a referenced `images/safari.png` file went missing, so I pulled it from the Safari Resources and copied it as a PNG.

No dependencies

## Checklist


aaronhmiller:
realized this is in the wrong directory structure! closing and will re-work it correctly

PR #532, opened 2021-08-03
adding force refresh script

aaronhmiller:
## Description

This script use the option-command-R keyboard shortcut to force refresh a Safari page. It arose out of a need to get a current CI badge status when using Github Action workflows because I noticed that the badge would not reflect the current status unless cache was cleared and the page refreshed. Normally that requires enabling Developer Menu and opt-cmd-E and cmd-R, but opt-cmd-R seems to do both and doesn't require enabling Developer Mode. Having this in Raycast is just one less thing to remember and it's nice to have it stored centrally.

## Type of change

- [X] New script command

## Screenshot

Not much to see here...it just force refreshes the frontmost safari page. I added a comment about this being undocumented and therefore subject to change w/o notice. 

none

## Checklist

### DIFF
--- a/commands/apps/safari/safari-clear-cache-reload.applescript
+++ b/commands/apps/safari/safari-clear-cache-reload.applescript
@@ -2,7 +2,7 @@
 
 # Required parameters:
 # @raycast.schemaVersion 1
-# @raycast.title Clear cache and refresh page
+# @raycast.title Clear Cache and Refresh Page
 # @raycast.mode silent
 #
 # Optional parameters:
--- a/commands/apps/safari/safari-clear-cache-reload.applescript
+++ b/commands/apps/safari/safari-clear-cache-reload.applescript
@@ -1,0 +1,18 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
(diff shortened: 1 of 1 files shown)


unnamedd (approved):
Looks good to me, thanks for your first contribution @aaronhmiller!

commands/apps/safari/safari-clear-cache-reload.applescript
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Clear cache and refresh page
```
fe9lix, line 5: Could you rename this to "Clear Cache and Refresh Page" to be consistent with Raycast commands?

aaronhmiller: Will do!

aaronhmiller:
anything else I should do to make this mergeable?

unnamedd:
@aaronhmiller, nope. If you have changed everything was asked, then just wait. As I'm not part of the team, I cannot merge it, but they would do it soon.

fe9lix (approved):

PR #539, opened 2021-08-07
adding system screenshot options

aaronhmiller:
## Description

The OS includes options to grab screenshots but remembering and executing the appropriate finger yoga is more difficult than using Raycast. There are options to save the fullscreen and selected area of the screen to both the Desktop and the Clipboard.

## Type of change

- [X] New script command

## Screenshot

Not much to see, just takes a screenshot.

no dependencies

## Checklist


es183923:
You should try using the `screencapture` cli. Some people may have remapped or disabled the default shortcuts (`⌘⇧3`, `⌘⇧5`, `⌘⇧5`).

aaronhmiller:
excellent suggestion @es183923! thank you. less dependency on something people might re-map as you say. and it feels a little faster now (perhaps b/c it's a shell script calling a binary instead of mapping some keystrokes?). anyway, happier w/ the results and changes. added some output confirmations for good measure. hardest part was mirroring the date format of the keyboard shortcut.

es183923:
You could do something like `screencapture -ui a`, so that it shows the UI when saving the image, instead of it directly saving to the desktop (the extra `a` isn't used, but only so that the command doesn't get upset that there isn't a location to save the file, even though it isn't used)

manpage: 
```
     -u      Present UI after screencapture is complete. Files passed to commandline will be ignored.
```

aaronhmiller:
Interesting idea. Maybe for a future iteration.

commands/system/capture-fullscreen-to-clipboard.applescript
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Capture Fullscreen to Clipboard
+# @raycast.mode silent
+# @raycast.packageName System
```
es183923: I recommend setting the package name to "Screenshot"

commands/system/capture-screen-selection-to-clipboard.applescript
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Capture Screen Selection to Clipboard
```
es183923: TO be consistent with the Cleanshot scripts, remove "Screen"

commands/system/capture-screen-selection-to-desktop.applescript
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Capture Screen Selection to Desktop
```
es183923: Remove "Screen" [`f7b9f72` (#539)](https://github.com/raycast/script-commands/pull/539/commits/f7b9f72aa32dda0fd0712e3e16c5120112f387c1#r684841634)

commands/system/capture-fullscreen-to-clipboard.applescript
aaronhmiller: I didn't feel comfortable creating a new package name, and for now System seems right.

commands/system/capture-screen-selection-to-clipboard.applescript
aaronhmiller: I think w/o "screen" someone might get mislead by other types of selections.

dehesa:
This is a great addition @aaronhmiller, thank you. Also props to @es183923 for the comments and reviews!

PR #575, opened 2021-09-02
Evaluate expression in Soulver

dlehman:
## Description

Open a new Soulver document and evaluate an expression.

## Type of change

- [X] New script command

## Screenshot

https://www.dropbox.com/s/ofbk7pit1qplzv1/soulver-eval-expression.mp4?dl=0

Soulver app must be installed (https://soulver.app/).

## Checklist


imthath-m:
Hey there, Thanks for the PR!

By the way, the in built calculator in Raycast is capable of doing the same and more. So, can you please explain what is the particular use case you are trying to solve.

CleanShot 2021-09-07 at 15 15 23

You can obviously use the script for personal needs, but I'm just trying to figure out what value does it add to other users.

dlehman:
Fair point and question.

Personally, I use Raycast built-in calculations for "one-offs" such as currency conversion, convert between two units, etc. I use Soulver when I need to do a longer set of calculations-- several calculations that depend on each other, or that I want to save to refer back to, or reuse. It's like a scratchpad for quick/temporary notes vs. starting a full document or spreadsheet.

But I totally see how it doesn't provide much functionality for most users, and doesn't need to be in the official repo. Thanks for your feedback.

imthath-m:
Thanks for understanding and explaining your use case. I'll close this PR.

PR #588, opened 2021-09-26
[Document fix] use workaround for header collision in Spotify

axblueblader:
## Description
- The link for Media>Spotify is wrongly pointing to App>Spotify
## Type of change
- [x] Documentation update
## Checklist

### DIFF
--- a/commands/README.md
+++ b/commands/README.md
@@ -101,7 +101,7 @@ This repository contains sample commands and documentation to write your own one
   - [Apple Music](#apple-music)
   - [Apple Tv](#apple-tv)
   - [Cmus](#cmus)
-  - [Spotify](#spotify)
+  - [Spotify](#spotify-1)
 - [Navigation](#navigation)
 - [Password Managers](#password-managers)
   - [Bitwarden](#bitwarden)


dehesa:
@unnamedd would you mind looking at the bot for a fix?

unnamedd:
Yeah, will take a look later into this issue!

unnamedd:
Hi @axblueblader and @dehesa,

I was checking now and in defense of our Toolkit, I need to argue that it is working correctly haha (sorry to write something you probably don’t want to read), the problem isn’t in the Toolkit itself, but in our way to add Spotify Script Commands in two different paths. When we generate the markdown, we do not check if that specific word was used before, that being said, we have two different places using the same anchor, then when you click on Spotify, you browser will redirect you for the first. 

With all that being said, my proposal is to fix that in a easier way for now which is moving the the unique Script Command for Spotify from Apps to the Spotify folder inside of Media. Just this small change, would fix that problem. 

I don’t think we should spend time trying to fix that in the Toolkit because it is an edge case and wasn’t rise up before and I don’t think it would happen again. 

@dehesa, would you mind to make the move of the Script Command to a folder where we have more Spotify Script Commands? 

If both you have any further question, please, let me know!

axblueblader:
Ok, moving the one command from Apps to Media sounds like a much better idea

dehesa:
That is a good simple solution :) I have already committed the changes to master.

PR #597, opened 2021-10-04
Added mood meter

Mitra98t:
Added: 
- script to add today's (or specified date's) mood
- script to view month mood table
- script to view year mood table
- file to store moods

## Description

Using mood-add.js the user can add a rating as his/her today's mood. The rating is from 1 to 5.
The script mood-month-display.js will output inline the monthly view.
The script mood-year-display.js will output all the month in one view (using fullOutput) the user can also specify the year to view, if the specified year is not present in the saved data the script will output `Empty year!`
moodTable.json is used to store the mood's ratings.

### Saving Data
The ratings are saved as object in arrays (representing months); those arrays are part of a bigger array (representing the year) all the years are saved as properties of the root object.
Every time one of all the script is run it checks if the current year is present in the root object. If it's not the script instantiate a new array structure for that year.

## Type of change

- [x] New script command

### Add Mood

Registrazione schermo 2021-10-04 alle 23 15 25

### Month Mood

Schermata 2021-10-04 alle 23 16 34

### Year Mood

Schermata 2021-10-04 alle 23 17 00

This scripts require nodejs to run

## Checklist


nhojb (changes requested):
Thanks for the PR - a really useful addition.

Could you make the following changes:
* Raycast uses "Verb Adjective/Noun" for commands so could you rename the mood commends to "Add Mood", "Display Mood Month" etc.
* It might be better to default to placing the "moodTable.json" in the user's home directory e.g. `~/.moodTable.json` - data should not be placed in the script's directory.
* The script fails if `moodTable.json` is not present. I would suggest a) do not include the json file in the repo b) the script should create the file if it does not exist.

PR #600, opened 2021-10-05
Add command to Decode QR code into the clipboard 

Dihgg:
## Description

This command uses [https://qrserver.com](https://qrserver.com) API to decode a screenshot of an **QR code**.

## Type of change
- [x] New script command

## Screenshot
Screen Recording 2021-10-05 at 4 00 10 PM

## Checklist


Dihgg:
> Why not use `jq`?

Don't want to add an dependency

nhojb (changes requested):
Thanks for the contribution.

Few issues:
1. Please use CamelCase for the command name i.e. "QR Code to Clipboard" (currently you are missing the leading cap on "clipboard").
2. It's hard to tell what to do after running the command - the macOS screen capture controls are quite hard to see. A message such as "Please capture the QR code" might be a useful prompt to the user.
3. It might also be useful to display a message after capture, but before processing has completed e.g. "Processing QR code..."
4. Character escaping of the QR code text doesn't seem correct. For example if I capture the QR code in the PR command and paste it into a text field e.g. Chrome's address field, I get `http:\/\/en.m.wikipedia.org/`.

nhojb:
Looks good. Thanks for the changes.

PR #619, opened 2021-11-07
Fix `audio-output-switch` for Monterey

mmerle:
## Description

It was brought to my attention by @dehesa that the `audio-output-switch` script I created was no longer working in macOS Monterey. From what I gathered it seems the error occurs due to the fact that Monterey now displays "Loading Sound..." for a moment while the preference pane is populated. Adding a simple 1 second delay to the script seems to fix the issue for me.

## Type of change

- [x] Bug fix

## Checklist


dehesa:
Thanks @mmerle that indeed solves the issue.

PR #622, opened 2021-11-11
Add git.io URL Shortener

astrit:
## Description

This is a very small script to shorten github.com urls from the official git.io url shortener when you copy the url.
I named this to git.io since it is a very niche url shortener and everyone who will use knows probably what it does.  

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

git-io

## Checklist


commands/browsing/git-io.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title git.io
```
unnamedd, line 5: I'm not quite sure about which kind of title we can use here, but git.io is definitely not good enough, based on the action that's going to be done, something like `Shortify URL` or `Shorten URL` and make the packageName `git.io` might be something better. What do you think?

astrit: I think It would make it confusing since it is not meant for general purpose only github.com and github.io links
And you can't really say `Shortify URL` or `Shorten URL`

These exist so not sure if it would confuse even more. 
- shorten-url-bitly.template.sh
- shorten-url.sh

The exact way would be `Shorten Github URL` but I guess a bit too long perhaps ?

unnamedd: > The exact way would be `Shorten Github URL` but I guess a bit too long perhaps ?

@astrit based for example in existing Script Commands, might be good:
`Shorten URL from GitHub`.

Check these examples in the screenshot:
image

ps: just be extra careful with the case of "H" in "Git**H**ub".

astrit: Should we go with `Shorten Github URL` or `Shorten URL from GitHub` what do you think ?

unnamedd: @astrit guess this one is better: `Shorten URL from GitHub`

astrit: @unnamedd I did rename to `Shorten URL from GitHub`

astrit:
@unnamedd check now if all good ✌️

unnamedd (approved):
Looks great to me, @astrit! 

Thanks a lot, man!

PR #626, opened 2021-11-19
GitHub contributions

astrit:
## Description
Show GitHub user contributions from the current year

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot
contributions

## Checklist


commands/dashboard/github-contributions.template..sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
unnamedd, line 1: Two and half points here worth to be mentioned:
1 - The file contains "two dots": `.template..sh`, we should have just one dot, i.e: `.template.sh`
2 - This Script Command will fit better together with others in: `developer-utils/github`
2.1 - Moving it to the folder mentioned, won't need to submit a logo as well, also, you will be able to use the dark version of the icon (which is already placed in the `images` folder there.

unnamedd (changes requested):

commands/dashboard/github-contributions.template..sh
astrit: The script is moved now but I did move the image and named same as the script since those images don't really fit if the light mode is going to be on so my icon covers both.

astrit:
@unnamedd — Updates done ✅

https://github.com/raycast/script-commands/pull/626#discussion_r753168428

commands/dashboard/github-contributions.template..sh
unnamedd: Actually they do, we have some other examples here that are using the same images, one for light and one for dark mode. It is better to reuse the resources already available to avoid duplications. 

The logo you moved here has one advantage which is be big, but in the case here, we don't need big images to act as icons, it is pretty okay to use the current ones. I will suggest the code, you just need to apply and it will be more conformant to the ones currently existing.

commands/developer-utils/github/github-contributions.template.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title GitHub Contributions
+# @raycast.mode inline
+# @raycast.refreshTime 2h
+
+# Optional parameters:
+# @raycast.packageName GitHub Contributions
+# @raycast.icon images/github-contributions.png
```
unnamedd, line 11:
```suggestion
# @raycast.icon images/github-logo.png
# @raycast.iconDark images/github-logo-iconDark.png
```

astrit:
@unnamedd thanks for the suggestion and I fixed it: 

```
# @raycast.icon images/github-logo.png
# @raycast.iconDark images/github-logo-iconDark.png
```

unnamedd (approved):
Now we are good to go! Thanks a lot for the changes and contributions, @astrit!

PR #629, opened 2021-12-07
Add script to place call via Apple's Cellular Calls feature

alexjsteffen:
Add a basic script using the native Cellular Call Feature provided by Apple. There is no easy way to make a normal telephone call other than using the Contacts app or FaceTime (although the UX for this feature has gotten worse in Monterey. 

See: https://support.apple.com/en-ca/HT209456 

## Description

This script uses the "iPhone Cellular Call Feature" provided by apple to quickly make telephone calls without having to go via Contacts. Added one second of sleep to allow an individual to consider whether it was a mistake. It happens.

## Type of change

- [ ] New script command

## Screenshot

As simple as can be:

Demo

None. 

## Checklist

- [ ] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md) - Yes

### DIFF
--- a/commands/communication/call-with-iphone.sh
+++ b/commands/communication/call-with-iphone.sh
@@ -11,7 +11,7 @@
 # @raycast.packageName Communication
 
 # Documentation:
-# @raycast.description Place a telephone call via your iPhone on wifi.
+# @raycast.description Place a telephone call via your iPhone on Wi-Fi.
 # @raycast.author Alexander JH Steffen
 # @raycast.authorURL https://github.com/alexjsteffen
 
--- a/commands/communication/call-with-iphone.sh
+++ b/commands/communication/call-with-iphone.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Call with iPhone
+# @raycast.mode compact
+
+# Optional parameters:
+# @raycast.icon 📱
+# @raycast.argument1 { "type": "text", "placeholder": "+1 514 555 1212" }
+# @raycast.packageName Call
+
+# Documentation:
+# @raycast.description Place a telephone call via your iPhone on wifi.
+# @raycast.author Alexander JH Steffen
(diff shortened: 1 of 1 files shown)


commands/communication/call-with-iphone.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Call with iPhone
+# @raycast.mode compact
+
+# Optional parameters:
+# @raycast.icon 📱
+# @raycast.argument1 { "type": "text", "placeholder": "+1 514 555 1212" }
+# @raycast.packageName Call
```
unnamedd, line 11:
```suggestion
# @raycast.packageName Communication
```

commands/communication/call-with-iphone.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Documentation:
+# @raycast.description Place a telephone call via your iPhone on wifi.
```
unnamedd, line 14:
```suggestion
# @raycast.description Place a telephone call via your iPhone on Wi-Fi.
```

unnamedd (changes requested):
Hey @alexjsteffen,

I made two really small suggestions in the code, nothing to worry about but it will make your Script Command more conformant to the others we have.

Thanks a lot for your first contribution on our community and welcome aboard!

commands/communication/call-with-iphone.sh
alexjsteffen: Good suggestions. Just merged your change. Cheers :)

unnamedd: @alexjsteffen would apply this suggestion here as well? After it we will be good to go! ;)

alexjsteffen:
Done. Have a great eve.

unnamedd (approved):
Now we are good to go! ✅ 
Thanks a lot for your contribution @alexjsteffen!

PR #643, opened 2022-01-03
Add microlink command

Kikobeats:
CleanShot 2022-01-03 at 13 03 01@2x

Hello,

This PR adds `microlink` command as an easy way to run [Microlink CLI](https://github.com/microlinkhq/cli) from Raycast.


commands/developer-utils/microlink.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -l
+
```
dehesa, line 2: `microlink` needs to be installed for this script to work, right? Could you please add a `# Dependency: ...` comment as described in [CONTRIBUTING.md](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md), please?

commands/developer-utils/microlink.sh
```diff
@@ -0,0 +1,23 @@
...
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Microlink API
+# @raycast.mode fullOutput
+
+# Optional parameters:
+# @raycast.icon https://cdn.microlink.io/logo/apple-touch-icon-60x60.png
```
dehesa, line 9: Also, add the icon in the `developers-utils/images` folder, make it 64x64 and just pass the icon name here.

dehesa (changes requested):

PR #660, opened 2022-01-22
Add a note about removing ".template"

hossammourad:
## Description

It took me some time to figure out that I need to remove the `.template` part of the name when I'm done with the script set up. This PR adds a simple comment in the script's file to remind users of the script to remove the `.template` part.

## Type of change

## Checklist

### DIFF
--- a/commands/system/audio/toggle-airpods.template.swift
+++ b/commands/system/audio/toggle-airpods.template.swift
@@ -16,6 +16,8 @@
 
 import IOBluetooth
 
+// Note: This a template which needs further setup. Remove `.template.` from the filename.
+
 // Get your device's MAC address by option (⌥) + clicking the bluetooth icon in the menu bar
 let deviceAddress = ""
 


unnamedd:
Hello @hossammourad, 
this is something described in the Troubleshooting section in our README. Even though I’m not against to have it in the Script Command itself, I agree this information needs to be even more clear, after all, we have tons of other Script Commands which are `.template` as well and we are not going to run one by one adding this note on them.

In any case, I’m going to copy the original author to this review, to see what he thinks before we move one.

cc @itsnwa

dehesa:
Hey @hossammourad ,

Thanks for giving feedback about this. As @unnamedd says we do have many scripts, so it wouldn't really scale to add a note to each of them and for future ones. However, I do agree that information could be better provided (moreover for people using Raycast the first time). That is why I added a small description and a link in the README's installation procedure (i.e. at the beginning of the document). Hopefully that helps.

PR #676, opened 2022-02-07
Add basic support for remote controlling Denon AVR's

Rediwed:
## Description

With this initial pull-request I’d like to add basic support for remote controlling modern Denon AV-receivers. The initial contribution contains six script commands:

1. configure-script-command.applescript can be used to (re-)configure your receiver’s IP-address. It also contains a method field for switching between classical method and modern method. However, currently **only modern method** is supported. I only have access to a modern receiver and am therefore not able to test older models. If anyone has access to a non-HEOS (pre-2016) Denon AVR, please let me know.
2. power-off.applescript powers off the Denon AVR
3. power-on.applescript powers on the Denon AVR
4. set-volume let’s you set the Denon AVR to a specific volume level (between 0 and 80)
5. volume-down.applescript decreases the volume by one unit
6. volume-up.applescript increases the volume by one unit

Each remote controlling script command (2 through 6) is able to read the ip-address configured by the configure script, as well as read the currently selected method (classic or modern).

## Type of change

- [x] New script commands

## Screenshot

Screenshot of Safari on 2022-02-07 at 12 02 27

Requirement: You need to have a ‘modern’ Denon AV-receiver. These are receivers that support HEOS, which is most receivers made after 2016. ‘Modern’ receivers can be identified by their model number. If it ends with an H, it’s HEOS compatible and works with this script command. I’ve begun work to support older receivers. If anyone can test the commands with me, they can be easily implemented in the dedicated ‘runClassic’ subroutines.

## Checklist


unnamedd (changes requested):
Hello @Rediwed, unfortunately we need you to submit the files in a readable way, not binary to make possible a code review.
I'd like to kindly ask you to copy the code into a readable file instead of submitting the content saved by the Script Editor.

As soon you submit it, we will be able to give a proper code review.
Thank you for your understanding.

PR #679, opened 2022-02-08
Add basic support for remote controlling Denon AVR receivers

Rediwed:
## Description

With this initial pull-request I’d like to add basic support for remote controlling modern Denon AV-receivers. The initial contribution contains six script commands:

- configure-script-command.applescript can be used to (re-)configure your receiver’s IP-address.

> It also contains a method field for switching between classical method and modern method. However, currently only modern method is supported. I only have access to a modern receiver and am therefore not able to test older models. If anyone has access to a non-HEOS (pre-2016) Denon AVR, please let me know.

- power-off.applescript powers off the Denon AVR
- power-on.applescript powers on the Denon AVR
- set-volume.applescript let’s you set the Denon AVR to a specific volume level (between 0 and 80)
- volume-down.applescript decreases the volume by one unit
- volume-up.applescript increases the volume by one unit

Each remote controlling script command (2 through 6) is able to read the ip-address configured by the configure script, as well as read the currently selected method (classic or modern).

(This time with plain text files)

## Type of change

- [x] New script command

## Screenshot

Screenshot of Finder on 2022-02-08 at 09 34 03

Requirement: You need to have a ‘modern’ Denon AV-receiver. These are receivers that support HEOS, which is most receivers made after 2016. ‘Modern’ receivers can be identified by their model number. If it ends with an H, it’s HEOS compatible and works with this script command. I’ve begun work to support older receivers. If anyone can test the commands with me, they can be easily implemented in the dedicated ‘runClassic’ subroutines.

## Checklist


dehesa:
Hi @Rediwed,

Great contribution! Some small details:
- Could you add the `@raycast.packageName Denon AVR` to all the scripts.

  This way all the scripts will have the same "subtitle" in the root search.

- It is recommended that all scripts related to the same product have the same icon.

  For example, all LG remote commands have the LG logo. I would recommend you do the same. Maybe a stylize D for "Denon" or the same "sound"-related emoji.

- The word "Denon AVR" is already in the subtitle (if you follow advice number 1). Thus, it would be repetitive to prefix all command's titles with `Denon AVR - `

Rediwed:
Hello @dehesa, thank you for you feedback.

I have made the following changes:
- Package name added to all scripts
- New images folder with icon for Denon AVR added, all scripts updated to include this icon
- Removed ‘Denon AVR - ‘ from script command names

Thanks!

dehesa:
One final thing:
- The icon description doesn't need the leading dot.

  ```
  // Turn this...
  # @raycast.icon ./images/logo.png
  // Into this...
  # @raycast.icon images/logo.png
  ```

- Name your icon something meaningful (e.g. denon.png).
  Most people have all their scripts in the same folder with a shared `images` subfolder.

Rediwed:
Thanks! I have performed the requested changes.

PR #706, opened 2022-03-27
Adding open last downloaded script

iloveitaly:
## Description

A simple script to open the last downloaded file.

Feature request: would be great if a modifier could be passed to a script so this could easily "view in finder" the last file instead of opening.

Worth noting that after I wrote this I found there was a swift script that did something similar. This shell script is ~10x faster and easier to understand, which is why I'm still submitting this PR.
## Type of change

- [x] New script command

## Screenshot

Hopefully it is simple enough this isn't needed.

N/A

## Checklist

### DIFF
--- a/commands/navigation/open-last-downloaded.sh
+++ b/commands/navigation/open-last-downloaded.sh
@@ -7,7 +7,7 @@
 # @raycast.packageName Navigation
 
 # Optional parameters:
-# @raycast.icon images/folder-home.png
+# @raycast.icon images/folder-downloads.png
 
 # Documentation:
 # @raycast.description Opens the last file that was downloaded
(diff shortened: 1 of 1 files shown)


commands/navigation/open-last-downloaded.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Open Last Downloaded
+# @raycast.mode silent
+# @raycast.packageName Navigation
+
+# Optional parameters:
+# @raycast.icon images/folder-home.png
```
dehesa, line 10: Wouldn't it be better if the icon is `folder-downloads.png`?

dehesa (changes requested):

dehesa:
Regarding your feature request: We do want to iterate on the scripts-commands engine. I cannot promise you when it will happen, but what you propose is very useful.

Regarding the Swift script: Swift scripts take a long time to compile the first time, afterwards they should be as fast as the shell scripts. In any case, it is much better to have the shell script. Thanks!

PR #722, opened 2022-05-14
Toggle WiFi - New command

pradeepb28:
## Description

This new command enables or disables wifi anytime

## Type of change

- [x] New script command

## Screenshot

https://user-images.githubusercontent.com/6823101/168410210-bfa1fb05-8d21-4e4c-b5d2-492523be85bf.mov

## Checklist


dehesa:
Hi @pradeepb28 ,

Thank you for the contribution, but how is this different from [this script](https://github.com/raycast/script-commands/blob/master/commands/system/wifi.template.applescript) which is already there?

unnamedd:
@dehesa unfortunately it is not different.

pradeepb28:
@dehesa I don't see 64px from this documentation https://developers.raycast.com/basics/prepare-an-extension-for-store#extension-icon

Regarding the script, I don't think the script is available to the end users I couldn't find it neither as an in-built nor in the store(there was other user who also couldn't find it, so he asked for this command in first place). I mean there is only one extension/command around Wi-Fi/WiFi which is different from what I've written :) 

Also the code similarity, 99% of applescript automations can be only implemented in one way only.

dehesa:
Hey @pradeepb28 ,

I have the feeling you are mixing script commands with extensions. Script commands (which this repo is all about) is related to user written scripts (usually in AppleScript, Swift, Python, etc.) performing a small functionality that need to be manually installed. Extensions are packaged functionality that you can easily install from the Raycast store (you can find extensions' source code [here](https://github.com/raycast/extensions) ).

As I pointed out in the previous message, there is already a script in this repo doing the same thing as your script. So there is no need to have two scripts doing the same thing. Also, the icon size is specified in this repo's README; the documentation you linked is about extensions.

pradeepb28:
@dehesa icon size has been fixed

>As I pointed out in the previous message, there is already a script in this repo doing the same thing as your script. So there is no need to have two scripts doing the same thing. Also, the icon size is specified in this repo's README; the documentation you linked is about extensions.

Regarding this like I've said it seems like it is not available on the user side

here are the results
CleanShot 2022-05-17 at 16 27 33@2x


CleanShot 2022-05-17 at 16 27 13@2x

dehesa:
> Regarding this like I've said it seems like it is not available on the user side.

Yes, script commands are not available from the store. Scripts commands are installed following [the steps written in the README](https://github.com/raycast/script-commands#install-script-commands-from-this-repository). Please read the README file and understand the difference between Script Commands, extensions, and quicklinks. Feel free to leave any questions you have about that.

pradeepb28:
I can't expect an end user would go and use those instructions to install a simple command. 

I will submit a new PR as an extension.

Thanks for your help 🙏

PR #732, opened 2022-06-06
Add command to search on Spotify Desktop App

marcoslor:
## Description

This PR adds a new AppleScript command that inputs a given query to the Spotify desktop app search field. 

## Type of change

- [x] New script command

## Screenshot

Preview

## Checklist

### DIFF
--- a/commands/media/spotify/search-spotify.applescript
+++ b/commands/media/spotify/search-spotify.applescript
@@ -16,45 +16,41 @@
 # @raycast.authorURL https://github.com/marcoslor
 
 on run argv
-
     try 
         if application "Spotify" is not running then
             activate application "Spotify"
             delay 1
-
-            set screenElementsCount to 0
-            set loopCount to 0
             
             # Verify whether the splash screen is gone by checking
             # if the count of the UI elements on the window is above a certain threshold.
 
-            repeat while (screenElementsCount < 6 and loopCount <= 20)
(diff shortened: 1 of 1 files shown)


marcoslor:
As this is actually my first PR to an open source project, having feedback is really important to me :)

dehesa:
Hi there @marcoslor ,

Thank you for the contribution. It looks you put some good work here. The script it is however an AppleScript relying on trial-and-error delays (which are finicky). Isn't the script functionality not already cover by the [Spotify Player extension](https://www.raycast.com/mattisssa/spotify-player) created by @mattisssa ? That extension relies on Spotify API and therefore is much more reliable.

Since quicklinks and extensions are supported by Raycast, we gently push people to use those instead of less-reliable AppleScripts. For example, all previous [Spotify media scripts](https://github.com/raycast/script-commands/tree/master/commands/media/spotify) have been softly deprecated and now we suggest everyone to use the [Spotify Control extension](https://www.raycast.com/thomas/spotify-controls).

marcoslor:
To who(m?)ever comes looking for something like this, it can be archieved with a quicklink: (`spotify://search/{query}`).
From my tests, it seems that Spotify accepts things like `album/id`, `artist/id`, etc.

I'd suggest the Raycast team adding it to the quicklink library.

Find in library dropdown

PR #734, opened 2022-06-14
add CopyCode script

Fatpandac:
## Description

Copy code from Chinese message.

## Type of change

- [X] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

image

## Checklist


dehesa:
Hi @Fatpandac,

Thank you for the contribution. Could you please explain further what this script does? Maybe send a gif/video showing how it works? I am not sure I am understanding it from the script code.

Fatpandac:
Just copy the code from a message in the Message APP.
cc

commands/apps/Message/copycode.sh
```diff
@@ -0,0 +1,33 @@
...
+# @raycast.authorURL https://github.com/Fatpandac
+
+USERNAME=$(id -u -n)
+
+result=$(sqlite3 /Users/$USERNAME/Library/Messages/chat.db 'SELECT text FROM message WHERE datetime(date/1000000000 + 978307200,"unixepoch","localtime") > datetime("now","localtime","-60 second") ORDER BY date DESC LIMIT 1;')
+
+keyword="验证码";
+
+if [ ! $result ]; then
+  echo "最近60秒未收到验证码！"
```
dehesa, line 23: I would appreciate if the `echo` is in English. Raycast doesn't currently support internationalization and English is (for now) the development language.

```
echo "No verification code received in the last 60 seconds!"
```

commands/apps/Message/copycode.sh
```diff
@@ -0,0 +1,33 @@
...
+  exit 0;
+fi
+
+if [[ "$result" =~ "$keyword" ]]; then
+  code=`echo $result | grep -o "[0-9]\{4,6\}"`;
+
+  echo "$code" | pbcopy;
+  echo "$code 验证码已复制"
```
dehesa, line 31: Same thing here:

```
echo "$code code copied!"
```

commands/apps/Message/copycode.sh
```diff
@@ -0,0 +1,33 @@
...
+# Documentation:
+# @raycast.description Copy code from chinese message.
+# @raycast.author Fatpandac
+# @raycast.authorURL https://github.com/Fatpandac
+
+USERNAME=$(id -u -n)
+
+result=$(sqlite3 /Users/$USERNAME/Library/Messages/chat.db 'SELECT text FROM message WHERE datetime(date/1000000000 + 978307200,"unixepoch","localtime") > datetime("now","localtime","-60 second") ORDER BY date DESC LIMIT 1;')
+
+keyword="验证码";
```
dehesa, line 20: I would make this an optional argument with a default value. You do that by setting the `@raycast.argument1` in the header:

```
# @raycast.argument1 { "type": "text", "placeholder": "code", "optional": true }
```

and then using `$1` for the `keyword`.

```
keyword=$1;
```

dehesa (changes requested):
Hi @Fatpandac 

Thank you for the gif and sorry for the delay reviewing this. The scripts is pretty neat and useful. I would recommend making this scrip "international" by letting the user select the word being search in the message. Currently you hardcode `验证码` but I believe it would be better if this would be an optional message which defaults in the code a user can hardcode. What do you think?

commands/apps/Message/copycode.sh
Fatpandac: Maybe I can create a keyword list, everyone can open other PR to add another keyword into the keyword list to support other language messages because the keyword just needs set up once.

dehesa: Sure, that seems good to me.

dehesa:
Thanks for the useful script @Fatpandac

PR #735, opened 2022-06-17
ocr_swift

wmszhe:
OCR with macOS Vision

### DIFF
--- a/commands/system/ocr.swift
+++ b/commands/system/ocr.swift
@@ -24,7 +24,8 @@ let screenCapturePath = "/tmp/ocr.png"
 let recognitionLanguages = ["en-US", "zh-CN"]
 let joiner = " "
 
-func screenCapture(_ command: String) throws -> String{
+@discardableResult
+func screenCapture(_ command: String) throws -> String {
     let task = Process()
     let pipe = Pipe()
 
@@ -36,7 +37,11 @@ func screenCapture(_ command: String) throws -> String{
     try task.run()
 
     let data = pipe.fileHandleForReading.readDataToEndOfFile()
-    return String(decoding: data, as: UTF8.self)
+    if let output = String(data: data, encoding: .utf8) {
+        return output
+    } else {
+        throw NSError(domain: "error", code: -1, userInfo: nil)
+    }
 }
 
 func convertCIImageToCGImage(inputImage: CIImage) -> CGImage? {
@@ -69,23 +74,24 @@ func recognizeTextHandler(request: VNRequest, error: Error?) {
 }
 
 func detectText(fileName: URL) {
-    if let ciImage = CIImage(contentsOf: fileName) {
-        guard let img = convertCIImageToCGImage(inputImage: ciImage) else {
-            return
-        }
+    guard
+            let ciImage = CIImage(contentsOf: fileName),
+            let img = convertCIImageToCGImage(inputImage: ciImage)
+    else {
+        return
+    }
 
-        let requestHandler = VNImageRequestHandler(cgImage: img)
+    let requestHandler = VNImageRequestHandler(cgImage: img)
 
-        // Create a new request to recognize text.
-        let request = VNRecognizeTextRequest(completionHandler: recognizeTextHandler)
-        request.recognitionLanguages = recognitionLanguages
+    // Create a new request to recognize text.
+    let request = VNRecognizeTextRequest(completionHandler: recognizeTextHandler)
+    request.recognitionLanguages = recognitionLanguages
 
-        do {
-            // Perform the text-recognition request.
-            try requestHandler.perform([request])
-        } catch {
-            print("Unable to perform the requests: \(error).")
-        }
+    do {
+        // Perform the text-recognition request.
+        try requestHandler.perform([request])
+    } catch {
+        print("Unable to perform the requests: \(error).")
     }
 }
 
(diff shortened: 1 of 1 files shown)


commands/system/ocr.swift
```diff
@@ -0,0 +1,139 @@
...
+
+func main() {
+    // Only support the system above macOS 10.15
+    guard #available(OSX 10.15, *) else {
+        print("Only support the system above macOS 10.15")
+        return
+    }
+    let _ = doShell("/usr/sbin/screencapture -i " + tmpPath)
+    guard recognitionQRCode(fileName: URL(fileURLWithPath: tmpPath)) else {
+        let _ = detectText(fileName: URL(fileURLWithPath: tmpPath))
```
dehesa, line 133: What is this function for? It is never returning anything

commands/system/ocr.swift
```diff
@@ -0,0 +1,139 @@
...
+// @raycast.mode silent
+
+// Optional parameters:
+// @raycast.icon 🖼
+// @raycast.packageName OCR_SWIFT
+
+// Documentation:
+// @raycast.author zhe
+// @raycast.authorURL https://github.com/wmszhe
+// @raycast.description OCR with macOS Vision
```
dehesa, line 15: I would appreciate a better description. What does this script do? Where is the input of the OCR and where the output goes?

dehesa (changes requested):
Hi @wmszhe 

Thank you for the contribution. I tried your code with a random QR code, but it doesn't work for me. Could you provide a QR code example with the PR, please?

Also, it would be great to have more detail on what the script actually do. Where does the script read the png from? Where does the output goes? I would appreciate if you code add a gif showing what the script does.

wmszhe:
Use macOS Vision API Identification pictures, if it contain a QR code, Copy the QR code content to the clipboard, If do not include QR codes, identify text content and supplement to clipboard

GIF: https://github.com/wmszhe/pichub/blob/3aad700130158a0a538504b82d8c1010c3d4bb3c/imgs/ocr.gif

commands/system/ocr.swift
```diff
@@ -0,0 +1,138 @@
...
+import Foundation
+import CoreImage
+import Cocoa
+import Vision
+
+let screenCapturePath = "/tmp/ocr.png"
+let recognitionLanguages = ["en-US", "zh-CN"]
+let joiner = " "
+
+func screenCapture(_ command: String) throws -> String{
```
unnamedd, line 27:
```suggestion
@discardableResult
func screenCapture(_ command: String) throws -> String {
```

commands/system/ocr.swift
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func main() {
+    // Only support the system above macOS 10.15
+    guard #available(OSX 10.15, *) else {
+        return print("Only support the system above macOS 10.15")
+    }
+    do {
+        // It must be ensured that the screenshot is preserved successfully
+        let _ = try screenCapture("/usr/sbin/screencapture -i " + screenCapturePath)
```
unnamedd, line 124: If you add the annotation `@discardableResult` in the `screenCapture` method, you won't need to make use of `let _ =`

```suggestion
        try screenCapture("/usr/sbin/screencapture -i " + screenCapturePath)
```

commands/system/ocr.swift
```diff
@@ -0,0 +1,138 @@
...
+        let request = VNRecognizeTextRequest(completionHandler: recognizeTextHandler)
+        request.recognitionLanguages = recognitionLanguages
+
+        do {
+            // Perform the text-recognition request.
+            try requestHandler.perform([request])
+        } catch {
+            print("Unable to perform the requests: \(error).")
+        }
+    }
```
unnamedd, line 89: You are unwrapping the content from `CIImage(contentOf:)` in the line 72, why not do the same with this guard? Actually, instead of a `if let` you can move the guard up and instead of a new level, you have the whole method statement at the same level.

```suggestion
    guard 
        let ciImage = CIImage(contentsOf: fileName), 
        let img = convertCIImageToCGImage(inputImage: ciImage) 
    else {
        return
    }

    let requestHandler = VNImageRequestHandler(cgImage: img)

    // Create a new request to recognize text.
    let request = VNRecognizeTextRequest(completionHandler: recognizeTextHandler)
    request.recognitionLanguages = recognitionLanguages

    do {
        // Perform the text-recognition request.
        try requestHandler.perform([request])
    } catch {
        print("Unable to perform the requests: \(error).")
    }
```

commands/system/ocr.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    task.launchPath = "/bin/bash"
+    task.arguments = ["-c", command]
+    task.standardOutput = pipe
+    task.standardError = pipe
+
+    try task.run()
+
+    let data = pipe.fileHandleForReading.readDataToEndOfFile()
+    return String(decoding: data, as: UTF8.self)
```
unnamedd, line 39: Question: is this method working? Because as far as I remember and I was checking here, `String(encoding:as)` isn't an existing method in the Foundation, the real one is `String(data:encoding)` which returns as an optional result.

That being said, I don't think this method here is working, if you convert to use the one I mentioned before, then it will be an optional and you need to make the method return and optional String (`-> String?`) or unwrap the content into a real String before return it here.

unnamedd (changes requested):
Hi @wmszhe,

thanks for your contribution to our repository, as we all Swift developers, we (me and Marcos) made some comments for you in order to make your contribution even better.

Check them out and let us what you think.

dehesa:
All seems good. Thank you for the work @wmszhe and sorry it took a bit of time to review this.

PR #736, opened 2022-06-23
Replace icon of toggle-sidecar command with an image as a workaround

jpuris:
## Description

The toggle-sidecar command using AppleScript does not run properly, if emoticon is used (in this particular case "🖥" is used). The script runs fine however, when run manually in Script Editor.

This change replaces the emoticon used for `@raycast.icon` with a png image.

This is merely a workaround. The root cause as to why this happens, is unknown to me.

## Type of change

- [ ] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

Error when running command with emoticon

error_with_emoticon

Icon (before)

before

Icon (after)

after

No change in requirements.

## Checklist


dehesa (changes requested):
Hey there @jpuris ,

Thank you for the feedback and the PR. Having a specific icon is better than having the emoji, but could you please make the icon 64x64 px?

I can shed some light on why the script is not running in Raycast, but it is running in the script editor (funnily we were talking exactly about that in [the Slack community](https://raycastcommunity.slack.com/archives/C028DTXTW3Y/p1655971789336899) yesterday).

Raycast is expecting the script to be encoded in one of the most popular formats (usually UTF-8). That is indeed what the majority of the text editors use. However, the Apple Script Editor uses a legacy encoding format that creates some problems when dealing with newer forms of text input such as emojis.

If you download the script from the repo, open it with any newish text editor (VSCode, TextEdit, etc.), modify the script, and then run it, Raycast won't have any problems. As soon as you open the script with the AppleScript Editor and save it, the encoding would be changed and Raycast will have trouble running it. We have tried to make Raycast to understand such legacy encoding, but it doesn't always work.

dehesa (approved):

PR #739, opened 2022-06-27
Local Weather

lukasoppermann:
## Description

This adds inline local weather to quickly check the current temperature.

It can be saved to the favorites to always see what temperature it is.

## Type of change

- [x] New script command

## Screenshot

Screenshot 2022-06-27 at 09 28 09

## Dependencies / Requirements

None

## Checklist

- [x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

PR #741, opened 2022-06-30
Updated current-weather to use better geo location

lukasoppermann:
As discussed on #739 this updates the geo location for the current weather script


dehesa:
That looks good to me @lukasoppermann, thank you. @crstauf, does it look good to you too?

crstauf:
No. That's four times as often, which means four times as many requests to wttr.in's data source, which means four times quicker to being throttled. Default should be one hour; if people change it, that's on them.

Yes, weather changes quickly, but a script command is not the place for real-time weather report. Plus, you can update on demand: no sense in updating it behind the scenes that frequently if it ain't being used.

PR #748, opened 2022-07-24
add Timing commands for starting/stopping timers

landendanyluk:
## Description

Adds support for Timing.app with two scripts for starting and stopping timers.

## Type of change

- [x] New script command

Requires Timing to be installed. Only works for users who have an Expert or Connect subscription, because the basic tier disables scripting support.

## Checklist


dehesa:
Looks great @landendanyluk. Thank you for the contribution.

PR #793, opened 2022-11-20
fix: warp commands fail sometimes due to the order of output

hthuong09:
## Description
- The `warp-cli status` command output order is not guaranteed so the script fails sometimes.
image

- Using `grep Status` to get the line with status instead of fixing to the second line.

## Type of change

- [ ] New script command
- [x] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Checklist


dehesa:
Thanks for the fix @hthuong09

PR #812, opened 2023-01-25
Added a script for getting battery level of bluetooth headphones

mSarheed:
## Description

Get's the battery level of bluetooth connected headphones

## Type of change

- [X] New script command

## Screenshot
bluetooth-headphones-battery-level

Nothing. Just add the Mac address of the bluetooth headphones

## Checklist


dehesa (changes requested):
Hey @mSarheed, thanks a lot for the contribution.

We already have a script for checking headphones battery, but it is specific for Airpods. Therefore, I am happy to merge this. I would only ask you to add the infix `.template.` to the filename since the script requires further tweaking by the users. For example `bluetooth-headphones-battery-level.template.sh`.

I know we haven't been consistent in requiring this in the past, but we are trying to be more consistent this days 😄

mSarheed:
@dehesa Should be fixed now :)

PR #820, opened 2023-02-16
Open Bing in Safari with Edge User-Agent

smxl:
Open Bing in Safari with Edge User-Agent, check Safari - Advanced - Develop first. This script works on the English Safari menu.

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Improvement of an existing script
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Checklist

- [x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

PR #821, opened 2023-02-17
Fix file encoding

xwjdsh:
## Description
I'd like to try #820, but when I added the `safari` folder as my script directory, I can't find it in the script commands.

image

After discovering that the issue was caused by the file encoding, I changed it to UTF-8, which resolved the problem.

image

## Type of change
- [x] Other (Specify)

## Checklist


dehesa:
Indeed. Thanks!

PR #859, opened 2023-07-23
Update github-repository-stars.template.rb

valdirjunior011:
## Description

Add loop for search for all repos just add owner info. and make sum for stars, forks, subscribers of all repos

## Type of change

## Screenshot

Repository Statistics

### DIFF
--- a/commands/developer-utils/github/github-repository-stars.template.rb
+++ b/commands/developer-utils/github/github-repository-stars.template.rb
@@ -1,10 +1,8 @@
 #!/usr/bin/env ruby
 
 # How to use this script?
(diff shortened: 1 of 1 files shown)


dehesa:
Hi @valdirjunior011,

Thank your for the contribution. I think you should let the previous script as it was and just create a new one with the functionality accumulating all repos. This way we can have both. What do you think?

Regards,
Marcos

valdirjunior011:
Hi @dehesa,
Sure! If you don't mind you can close this PR. I will make a new one to add the new script.

sounds good ?

PR #869, opened 2023-08-16
feat: add open icloud

DannySeidel:
## Type of change

- [x] New script command

## Screenshot

NA
opens icloud folder in finder

no special requirements

## Checklist


dehesa:
Hi there @DannySeidel,

Thank you for the contribution. However, since Raycast introduced [Quicklinks](https://manual.raycast.com/quicklinks), we have stopped accepting these type of simple scripts. I would recommend you to check the Quicklinks out.

DannySeidel:
Hi @dehesa,

thanks for the suggestion I did not know that. Maybe you could remove all simple scripts and provide them as Quicklinks via a explore command like snippets or prompts. This would probably introduce users to the possibilities of Quicklinks and minimize confusion.

dehesa:
Yeah, we thought about it. We are keeping the simple scripts for historical reasons. That is, many people put a lot of effort at the time and we wanted them to be "seen". But maybe it is time to start deleting them.

Regarding a Quicklink explorer, we don't have anything like that. Quicklink are suppose to be something extremely simple that any one can pick up. Not sure, it is worth it.

PR #899, opened 2024-01-06
Adds current browser check for compatibility with Chrome confirmation…

scottaw66:
… dialog when switching from Chrome to another browser

## Description

I modified the following three files in commands/system:

default-browser-arc.applescript
default-browser-firefox.applescript
default-browser-safari.applescript

The changes are identical in each:

I added a check for the current default browser. IF the current default browser is Chrome, the AppleScript needs to click button 1 of the confirmation dialog that appears when defaultbrowser is called (this is a change), otherwise it needs to click button 2 (the current action).

The reason for this is a difference in the order of the buttons on the dialog that appears if Chrome is the current default browser vs. the order if any other browser is default. Chrome's dialog has "change" as button 1 and "keep" as button 2, while the dialog for other browsers has them in the reverse order.

NOTE: I did not check Chromium so I don't know if this also applies to the case when Chromium is the default.

## Type of change

- [ ] New script command
- [x ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Checklist

- [x ] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

### DIFF
--- a/commands/system/default-browser-arc.applescript
+++ b/commands/system/default-browser-arc.applescript
@@ -17,6 +17,9 @@
 # @raycast.authorURL https://github.com/dehesa
 # @raycast.description Set Arc as the default browser.
 
+# check to see what the current browser is
+set currentDefaultBrowser to my getCurrentDefaultBrowser()
+
 set repeatCount to 0
 
 tell application "System Events"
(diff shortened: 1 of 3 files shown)


dehesa:
Hey @scottaw66, thank you for the changes. It works!

PR #903, opened 2024-01-23
Add script to delete line at current cursor position

annietma:
## Description

This script deletes the line at the current cursor position, which is helpful when writing code.

## Type of change

- [x] New script command

## Screenshot

CleanShot 2024-01-23 at 14 43 54

## Checklist

- [ ✔️] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)


commands/developer-utils/delete-current-line.applescript
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Delete Current Line
+# @raycast.mode silent
```
dehesa, line 6: Please add under this line:

```
# @raycast.packageName Developer Utilities
```

commands/developer-utils/delete-current-line.applescript
```diff
@@ -0,0 +1,19 @@
...
+# Optional parameters:
+# @raycast.icon ⌨️
+
+# Documentation:
+# @raycast.description This script deletes the line at cursor position.
+# @raycast.author Annie Ma
+# @raycast.authorURL http://www.anniema.co/
+
+tell application "System Events"
+    keystroke "k" using control down
```
dehesa, line 17: I think it would be best to use the ⌘+DEL

dehesa (changes requested):
Hi @annietma,

Thank you for the contribution. I didn't know about the '⌃+K` keyboard shortcut. Neat! That shortcut works with VSCode (and some other specialized editors), but it doesn't work anywhere else (e.g. Apple Notes, TextEdit, even Xcode). I think a better alternative to `⌃+K` is `⌘+DEL`, which does the same thing and it works everywhere. What do you think?

commands/developer-utils/delete-current-line.applescript
annietma: ⌘+DEL only deletes everything to the left of the cursor, so if the cursor is not at the end of the line, it will only partially delete the line. I used ^ + K in combo with ⌘+DEL because it's a [default Apple shortcut](https://support.apple.com/en-us/HT201236) that deletes everything to the right of the cursor. This shortcut works for me everywhere, but if you'd prefer, I'm happy to file it under a specific app like VSCode instead of general developer shortcuts.

dehesa: Sorry for the confusion, I meant to still do a combo:
1. `⌘+DEL` (which does the same as `⌃+K`).
2. `⌘+⌫` (which you currently have with `keystroke (ASCII character 8) using command down`).

Basically I am advocating to replace `⌃+K` by `⌘+DEL` since it works in most apps, while `⌃+K` only work in a very small subset of ups.

Please notice `DEL` and `⌫` are not the same key. One of them deletes forward while the other deletes backwards.

> This shortcut works for me everywhere.

Does it? If you open Apple Notes (for example). Does the `⌃+K` works for you? It doesn't for me.

annietma: Sorry, I didn't catch the distinction between delete and forward delete! However, ⌘+DEL (forward delete) doesn't work for me anywhere; only ^+K does. Are you using the default Apple keyboard configurations? According to the [official documentation](https://support.apple.com/en-us/HT201236), the keyboard shortcut to delete the rest of the line/paragraph to the right of the cursor is ^+K, not ⌘+DEL:
"Control-K: Delete the text between the insertion point and the end of the line or paragraph."

Here is a demonstration of ^+K in Apple Notes:
CleanShot 2024-01-25 at 23 17 07

Here is a demonstration of my shortcut in Apple Notes:
CleanShot 2024-01-25 at 23 21 13

If it still isn't working for you, I'm happy to file this under VSCode or a specific text editor application where it 100% works for both of us.

commands/developer-utils/delete-current-line.applescript
annietma: Added!

commands/developer-utils/delete-current-line.applescript
dehesa: > Are you using the default Apple keyboard configurations?

I do have a different keyboard and a different Keyboard layout. Besides I am in a European locale. So, that must probably be the issue.

Sorry for the back and forth. I am happy to have learn about `⌃+K`; thank you!

annietma: No problem, thanks for approving😊

PR #908, opened 2024-02-19
update the script API for apple music love song action to match current properities

hugivar:
## Description

Apple Music has seemingly changed their properities data since the time of writing this script. Found this useful util here to gather all the available information that could be modified

```
tell application "Music" to get properties of current track
```

## Type of change

- [x] Bug fix

## Checklist

### DIFF
--- a/commands/media/apple-music/apple-music-love-current-track.applescript
+++ b/commands/media/apple-music/apple-music-love-current-track.applescript
@@ -13,7 +13,7 @@
   tell application "Music"
     if player state is playing then
         set favorited of current track to true
-        log "<3 Yeah, that song is rad."
+        log "<3 Yeah, the song is now favorited"
     else
         return "No track is currently playing."
     end if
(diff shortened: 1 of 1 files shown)


commands/media/apple-music/apple-music-love-current-track.applescript
```diff
@@ -11,6 +11,10 @@
 # @raycast.schemaVersion 1
 
   tell application "Music"
-    set loved of current track to true
-    log "<3 Yeah, that song is rad."
+    if player state is playing then
+        set favorited of current track to true
+        log "<3 Yeah, that song is rad."
```
unnamedd, line 16: > "<3 Yeah, that song is rad." 
regarding the message here, I guess we can go a little bit different and, instead of saying that, we answer with a another kind of message, something like: "<3 Yeah, the song is now loved/favorited" something on that way, what do you think?

hugivar: Works for me. Updated!

PR #916, opened 2024-03-12
[DevUtil - create-image-from-code] Simplify title encoding logic and change the params to pass correctly

teepobharu:
## Description

- fix encoding to use from the args function itself ( no need for external deps  - php)
- move the params to be after hash part so the website reads the parameter correctly
    - Params not sent - [ray.so?title=title](ray.so?title=title)
    - Params sent correctly -  [ray.so#title=title](ray.so#title=title)

## Type of change

### Title With spaces 
https://ray.so/?#title=title%20with%20space&code=bGFuZ3VhZ2U9JExBTkdVQUdF&colors=midnight&background=true&darkMode=true&padding=64&title=title%20with%20space&code=bGFuZ3VhZ2U9JExBTkdVQUdF&language=auto

image

## Checklist

- [ ] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)


unnamedd:
Hi @teepobharu,

Just to let you know that last week, I reached out to our friends from Raycast about the problem you found. They should respond soon, and I will inform you whether to proceed with the review of your PR or close it.

To explain why I initiated this process:
The issue you are addressing regarding parameter injection, converting from `?` to `#`, is a bug in some internal implementation of ray.so, possibly a regression. That's why they need to conduct an internal check before we proceed here.

But don't worry, you haven't been forgotten here! 😊

samuelkraft:
Hi @teepobharu, thanks for the PR! 
The way we handle params was changed a while back from `?` to `#` as you correctly identified, so this is the expected format now. 

Cheers, Samuel

PR #841, opened 2023-05-18
feat: convert raw html on clipboard to rich text

iloveitaly:
## Description
This script will convert raw HTML on your clipboard to rich text.

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Improvement of an existing script
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot
NA

## Dependencies / Requirements
Requires pandoc

## Checklist

- [x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

PR #685, opened 2022-02-16
Paste as plain text

iloveitaly:
## Description

Coming from Alfred, this was one of the things I was missing. This is a simple
script that can be invoked to paste a string without any formatting.

## Type of change

- [x] New script command

## Screenshot

https://user-images.githubusercontent.com/150855/154361647-7d153c9b-141b-4237-b8ee-b400ccf146b4.mp4

## Dependencies / Requirements

N/A

## Checklist

- [x] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

PR #917, opened 2024-03-12
feat: adding show-notifications command

iloveitaly:
## Description

Applescript to expand notifications on macOS

## Type of change

- [x] New script command

## Screenshot

Run the script, notifications are expanded

None, `osascript` is bundled with macOS

## Checklist

### DIFF
--- a/commands/system/show-notifications.applescript
+++ b/commands/system/show-notifications.applescript
@@ -1,0 +1,25 @@
+#!/usr/bin/osascript
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Show Notifications
+# @raycast.mode silent
+
+# Optional parameters:
+# @raycast.icon 🔕
+# @raycast.packageName System
+
+# Documentation:
+# @raycast.description Show all notification of the first alert group
+# @raycast.author iloveitaly
+# @raycast.authorURL github.com/iloveitaly
+
+tell application "System Events"
+	tell process "NotificationCenter"
+		if not (window "Notification Center" exists) then return
+		-- clicking on the notification group will expand it and show all sub messages
+		click of group 1 of UI element of scroll area of first group of window "Notification Center"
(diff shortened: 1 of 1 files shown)


unnamedd (approved):
It looks good to me, however, as I am using British English, it doesn't work for me or to anyone else who isn't using American English in the System, perhaps you want to write something to warn the user about that.

iloveitaly:
How does it not work? What word is problematic?

unnamedd:
> How does it not work? What word is problematic?

I use my system in the British English, which means that the word "_Center_" (US), for me, and other users, will be "_Centre_" (UK). As this is the case of your Script Command, worth to mention such problem in body of the file, because I am quite sure someone may download it and it will not work, but the user won't understand why. 

At least explaining that, it won't solve, but will explain.

Let me show you how the system identify it for me.
image

image

grzegorzkrukowski:
hey @iloveitaly for me even in US English the script is not working. It doesn't detect "Notification Center" window until I expand it manually - only then the script picks up the window and executes the `click of the group` which is probably not the desired behavior.

Maybe we can find a better way to find a window than using the localized name? 
Can the Notification Center process have more windows? If not maybe we can always pick the first one? 

As @unnamedd mentioned - would be great to add error handling in case the window cannot be found and we cannot find a reliable way to make it work for all languages.

dehesa:
Hey @iloveitaly,

This is another PR where you don't react on feedback (like #685 or #841). We do appreciate the contributions, but it is annoying to keep the PRs open for such a long time.

I will be closing this one for now.

iloveitaly:
@dehesa understand it's annoying, but being ok with longer feedback cycles when a free open-source contribution is being offered is a good approach to encouraging more contributions.

Is the stance here that if a script does not support all locales it shouldn't be accepted? In this case I can't test the UK wording easily and someone could easily mutate this script to adjust the wording.

PR #934, opened 2024-04-25
Fix: Natural Scrolling Script for latest macos versions

koiralakiran1:
## Description

The current natural scrolling script doesn't work in latest macos version 14 (Sonama). I think this has been broken for a while (after Ventura), based on my research, but I can't confirm it. This change fixes the script to work correctly for macos 14 (Sonama) and higher and keeps the old script for older versions. I think the same new script would work for older versions as well, but I can't test that, so I've kept the old script under the else block.

## Type of change

- [ ] New script command
- [x] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Checklist


dehesa:
Thank you for the fix @koiralakiran1 🎉

PR #945, opened 2024-06-22
Fix: ocr command by updating recognitionlangues params

sivagao:
## Description

Previously, Chinese could not be recognized by OCR, it was all garbled.
Correct the OCR command by updating the 'recognition languages' parameters; this will ensure both Chinese and English are recognized accurately.

Reference:
To recognize traditional and simplified Chinese, specify zh-Hant and zh-Hans as the first elements in the request’s recognitionLanguages property. English is the only other language that you can pair with Chinese. https://developer.apple.com/documentation/vision/original_objective-c_and_swift_api/recognizing_text_in_images

## Type of change

- [ ] Bug fix

## Checklist

### DIFF
--- a/commands/system/ocr.swift
+++ b/commands/system/ocr.swift
@@ -21,7 +21,7 @@ import Cocoa
 import Vision
 
 let screenCapturePath = "/tmp/ocr.png"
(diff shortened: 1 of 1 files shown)


dehesa:
Thank you for the changes @sivagao

PR #946, opened 2024-06-22
Remove superfluous period from all script template renaming instructions

dmatt:
## Description

In the instruction comments for each script template there is a comment that says:

`// Duplicate this file and remove ".template." from the filename to get started.`

The full file path is, for example, `../raycast scripts/script-command.template.sh` so following this instruction as is would result (undesirebly) in:

`../raycast scripts/script-commandsh`

which isn't a valid file. So the instruction should be:

`// Duplicate this file and remove ".template" from the filename to get started.`

The .js file templates/script-command.template.js did not have this issue so it is not touched in this commit/PR.

## Type of change

- [ ] New script command
- [ ] Bug fix
- [X] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

n/a

n/a

## Checklist

### DIFF
--- a/templates/script-command.template.applescript
+++ b/templates/script-command.template.applescript
@@ -2,7 +2,7 @@
 
 # Raycast Script Command Template
 #
-# Duplicate this file and remove ".template." from the filename to get started.
+# Duplicate this file and remove ".template" from the filename to get started.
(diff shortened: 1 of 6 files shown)


dehesa:
You are very correct. Thank you!

PR #954, opened 2024-08-07
Add script to open Claude and submit prompt

nimobeeren:
## Description

This script opens Claude in Chrome browser and submits the prompt provided as argument.

## Type of change

New script command

## Screenshot

https://github.com/user-attachments/assets/49f21e5f-5574-44c5-a559-4719ca94db28

This script requires `chrome-cli` installed: https://github.com/prasmussen/chrome-cli

Install via homebrew: `brew install chrome-cli`

## Checklist

### DIFF
--- a/commands/apps/claude/claude.js
+++ b/commands/apps/claude/claude.js
@@ -1,0 +1,120 @@
+#!/usr/bin/env node
+
+// Dependency: This script requires `chrome-cli` installed: https://github.com/prasmussen/chrome-cli
+// Install via homebrew: `brew install chrome-cli`
+
+// Required parameters:
+// @raycast.schemaVersion 1
+// @raycast.title Prompt Claude
+// @raycast.mode silent
+// @raycast.packageName Claude
+
+// Optional parameters:
+// @raycast.icon 🧠
+// @raycast.argument1 { "type": "text", "placeholder": "prompt"}
+
+// Documentation:
+// @raycast.description Open Claude in Chrome browser and submit a prompt
+// @raycast.author Nimo Beeren
+// @raycast.authorURL https://github.com/nimobeeren
+
+const { execSync } = require("child_process");
+
+const prompt = process.argv[2];
+
+process.env.OUTPUT_FORMAT = "json";
+
+/** Escape a string so that it can be used in JavaScript code when wrapped in double quotes. */
+function escapeJsString(str) {
+  return str.replaceAll(`\\`, `\\\\`).replaceAll(`"`, `\\"`);
+}
+
+/** Escape a string so that it can be used in a shell command when wrapped in single quotes. */
+function escapeShellString(str) {
+  return str.replaceAll(`'`, `'"'"'`);
+}
+
+try {
+  execSync("which chrome-cli");
+} catch {
+  console.error(
+    "chrome-cli is required to run this script (https://github.com/prasmussen/chrome-cli)"
+  );
+  process.exit(1);
+}
+
+// Find the Claude tab if one is already open
+let tabs = JSON.parse(execSync("chrome-cli list tabs")).tabs;
+let claudeTab = tabs.find((tab) => tab.url.startsWith("https://claude.ai/"));
+
+// If there is a Claude tab open, get its info. Otherwise, open Claude in a new
+// window.
+let claudeTabInfo;
+if (claudeTab) {
+  // Open a new Claude session in the existing tab and focus it
+  execSync(`chrome-cli open 'https://claude.ai/new' -t ${claudeTab.id}`);
+  // Get tab info
+  claudeTabInfo = JSON.parse(execSync(`chrome-cli info -t ${claudeTab.id}`));
+} else {
+  // Open a Claude session in a new window, focus it and return the tab info
+  claudeTabInfo = JSON.parse(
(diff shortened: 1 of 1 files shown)


commands/apps/claude/claude.js
```diff
@@ -0,0 +1,120 @@
...
+
+// Required parameters:
+// @raycast.schemaVersion 1
+// @raycast.title Prompt Claude
+// @raycast.mode silent
+// @raycast.packageName Claude
+
+// Optional parameters:
+// @raycast.icon 🧠
+// @raycast.argument1 { "type": "text", "placeholder": "prompt"}
```
unnamedd, line 14: The placeholders need to be camel case

```suggestion
// @raycast.argument1 { "type": "text", "placeholder": "Prompt"}
```

commands/apps/claude/claude.js
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env node
+
+// Dependency: This script requires `chrome-cli` installed: https://github.com/prasmussen/chrome-cli
+// Install via homebrew: `brew install chrome-cli`
```
unnamedd, line 4: Would be better if you note here that `node` is another requirement, since `node` isn't something pre-installed on macOS.

commands/apps/claude/claude.js
```diff
@@ -0,0 +1,120 @@
...
+
+// Required parameters:
+// @raycast.schemaVersion 1
+// @raycast.title Prompt Claude
```
unnamedd, line 8: Just an idea, but what do you think about
```suggestion
// @raycast.title Ask Claude
```
?

commands/apps/claude/claude.js
```diff
@@ -0,0 +1,120 @@
...
+/** Escape a string so that it can be used in a shell command when wrapped in single quotes. */
+function escapeShellString(str) {
+  return str.replaceAll(`'`, `'"'"'`);
+}
+
+try {
+  execSync("which chrome-cli");
+} catch {
+  console.error(
+    "chrome-cli is required to run this script (https://github.com/prasmussen/chrome-cli)"
```
unnamedd, line 41: This means your Script Command will only work on Chrome, maybe you should open the usage of your Script Command to users of other browsers, even though, many of them uses Chromium as engine, sounds to me that `chrome-cli` will work only for Chrome.

Is something that important to be able to search for an open Claude tab in the browser to reuse it with a new prompt?

unnamedd (changes requested):
Hi @nimobeeren, thanks for you first contribution to our repository,

I've made some small comments on your PR, give a check and re-request a review when you finish to address them.

commands/apps/claude/claude.js
nimobeeren: `chrome-cli` indeed works with other Chromium-based browsers through aliases such as `brave-cli` (see [the `chrome-cli` readme](https://github.com/prasmussen/chrome-cli/tree/master?tab=readme-ov-file#other-browsers)). However, I'm not sure how to detect which browser the user would prefer. The most straight-forward approach I see is letting the user modify this script if they want to use another browser.

I could add a variable at the top of the script to make it easier to change the `chrome-cli` executable. What do you think is best?

commands/apps/claude/claude.js
nimobeeren: done

nimobeeren:
@unnamedd thanks for your review! I've implemented or replied to your comments.

I've also changed behavior to open in a new tab instead of new window, and added a note to tell the user to enable JavaScript from Apple Events. Could you give it another look?

commands/apps/claude/claude.js
unnamedd: Nice @nimobeeren!

Reading it now, it looks much simpler, which led me to another question (and this is not a blocker, but something for you to consider): would it not be better if you convert from Node to a Shell script? Since you are just checking whether `chrome-cli` is installed or not, perhaps you can do all that process in Shell, avoiding even the need for a Node installation.

unnamedd: > I could add a variable at the top of the script to make it easier to change the `chrome-cli` executable. What do you think is best?

Maybe in this case, you could make the extension a template (just adding `.template` in the file name and tell the user to uncomment the line related to its browser: brave, chrome and so on. What do you think?

unnamedd (approved):
Hey @nimobeeren,

thanks for you contribution! Even though there are something we could improve here, it looks good to me! 

If you prefer to make improvements later, will be a pleasure to review them later!

unnamedd:
@nimobeeren I found this link earlier today ([Claude’s API now supports CORS requests, enabling client-side applications](https://simonwillison.net/2024/Aug/23/anthropic-dangerous-direct-browser-access/)). Perhaps it might be interesting to you.

nimobeeren:
> @nimobeeren I found this link earlier today ([Claude’s API now supports CORS requests, enabling client-side applications](https://simonwillison.net/2024/Aug/23/anthropic-dangerous-direct-browser-access/)). Perhaps it might be interesting to you.

Thanks, that's good to know! I think this only applies to the API though, rather than the user-facing Claude product. I think there is a place for API-based interactions with LMMs (like Raycast AI), but this script is meant to work with the (free or paid) web interface.

PR #956, opened 2024-08-13
Hide Application

that70schris:
## Description

This command hides the frontmost app.  This is useful if you want to remap the default Hide menu command keyboard shortcut: `command + h`.

## Type of change

- [x] New script command

## Checklist

### DIFF
--- a/commands/navigation/hide-application.applescript
+++ b/commands/navigation/hide-application.applescript
@@ -1,0 +1,18 @@
+#!/usr/bin/osascript
+
+# @raycast.schemaVersion 1
+# @raycast.author Chris Bailey
+# @raycast.authorURL https://raycast.com/that70schris
+# @raycast.description Easily hide your foremost application
+# @raycast.packageName Navigation
+# @raycast.title Hide Current Application
+# @raycast.mode silent
+# @raycast.icon 🫥
+
+tell application "System Events"
+	tell (first process whose frontmost is true)
+		set _name to displayed name
+	end tell
+
+  set visible of application process _name to false
+end tell


unnamedd (commented):
Hi @that70schris, thanks for you contribution.

However, I do think we have two problems.

First, your Script Command assumes the user does have Swift installed and this isn't true (not going to mention that is missing the indication that Swift needs to be installed).

Second, macOS does provide the possibility to hide any frontmost apps just by pressing Command + H, this is system-wide shortcut as you can see in the first screenshot.
Also, still in the second point, Raycast also provides a way to hide an application, give a check in the second screenshot. 

Then, I'd like to raise the question: what would be the benefit of this Script Command you are proposing?

First screenshot
image

Second screenshot
image

that70schris:
@unnamedd ok, I originally had it as an AppleScript so I can revert it back to that.  

to the second point, I am aware that `command + h` is available system wide and in Raycast.  However, there is no way to remap this shortcut system wide as there is with shortcuts for other menu commands.  as I stated in my description, this command is to specifically to allow the remapping of the keyboard shortcut

that70schris:
Also, as far as needing an indication that swift is required, I'm not sure what you mean.  Looking through other swift commands, I don't see an example of this

unnamedd:
> Also, as far as needing an indication that swift is required, I'm not sure what you mean. Looking through other swift commands, I don't see an example of this

In the section [Scripts that require installation of runtimes and dependencies](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#scripts-that-require-installation-of-runtimes-and-dependencies) of our CONTRIBUTING.md you will be able to see it. 😉

> to the second point, I am aware that command + h is available system wide and in Raycast. However, there is no way to remap this shortcut system wide as there is with shortcuts for other menu commands. As I stated in my description, this command is to specifically to allow the remapping of the keyboard shortcut

It makes sense for you. Maybe it won't make sense for others (myself included), especially because the shortcut is available system-wide, but I am not here to judge; I am here to raise points.

> The way I personally use this is to first turn my `capslock` key into an `esc` key, and then assign `command + esc` to "Hide Application". At least while I'm still in the habit of using `alt + tab` to navigate, this addition makes it really easy to keep my thumb on the `command` key, and just move my pinky from `tab` to `capslock`

Just one thing regarding this remap: Raycast uses Command + ESC to return to the root search. This is very handy when you are deep inside Raycast (extensions, details, more details about something). You can use that shortcut to return instead of just closing Raycast.

Last but not least: it's very interesting that you had it as an Apple Script and then reverted it. This way, there's no need to install anything, making it even easier to use your Script Command.

Thanks for the revert and for the explanations.

commands/navigation/hide-application.applescript
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/osascript
+
+# @raycast.schemaVersion 1
+# @raycast.author that70schris
```
unnamedd, line 4:
```suggestion
# @raycast.author Chris Bailey
```

commands/navigation/hide-application.applescript
```diff
@@ -0,0 +1,17 @@
...
+# @raycast.authorURL https://raycast.com/that70schris
+# @raycast.packageName Navigation
+# @raycast.title Hide Application
```
unnamedd, line 7: I guess it can be "Foremost", "Frontmost" or "Current". What do you think?

```suggestion
# @raycast.title Hide Current Application
```

commands/navigation/hide-application.applescript
```diff
@@ -0,0 +1,17 @@
...
+# @raycast.mode silent
+# @raycast.icon 🫥
+
```
unnamedd, line 10: Would be interesting to add a description to your Script Command to make it more beautiful in the Awesome Script Command and also, inside of Raycast.

```suggestion
# @raycast.description Easily hide your foremost application

```

unnamedd (changes requested):

that70schris:
yea, I was actually trying to find a way to prevent this script from hiding the Raycast prompt, do you know if that's possible?

unnamedd:
I don't think it will be possible to override it 😔

that70schris:
Instead of overriding it I was looking for a way to detect that the Raycast prompt was open and to skip "set visible to false"

unnamedd:
I guess that would only be possible if Raycast had an API for AppleScript, but as far as I know, they do not. I am quite sure that won't be easy to achieve.

that70schris:
thanks @unnamedd, I've made the updates you suggested

unnamedd (approved):
@that70schris thanks for the changes and your contribution to our repository.

It looks good to me! ✅

PR #975, opened 2024-12-16
Add a new script of rename-videos-pictures.py

StepaniaH:
## Description

A recursive file renaming script that helps batch rename video and image files with a consistent date-based pattern. The script renames files to `foldername-MMDD-NN` format (e.g., `videos-1216-01.mp4`), where:
- `foldername`: name of the containing folder
- `MMDD`: current date in month-day format
- `NN`: sequential number starting from 01

Features:
- Recursively processes all subdirectories
- Supports common video and image formats
- Maintains existing renamed files
- Provides dry-run mode for preview
- Allows filtering by file type (video/image)

## Type of change

- [x] New script command

## Screenshot

rename-vedios-pictures.py-gif

No additional dependencies required. The script uses only Python standard library.

## Checklist

### DIFF
--- a/commands/system/rename-vedios-pictures.py
+++ b/commands/system/rename-vedios-pictures.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Rename Video
+# @raycast.mode fullOutput
+
+# Optional parameters:
+# @raycast.icon 📂
+# @raycast.argument1 { "type": "text", "placeholder": "Placeholder" }
+
+# Documentation:
+# @raycast.description This is a simple Python script for recursively renaming video and picture files within a directory. Type the root directory's absolute path, and it will scan all the video and picture files in it and rename them according to the folder where they are located as the format `<folder_name>-<current_date (MMDD)>-<incremental_number>`.
+# @raycast.author StepaniaH
+# @raycast.authorURL https://github.com/StepaniaH
+
+import os
+import sys
+import datetime
+import argparse
+import re
+
+class RenameFilesAsDate:
+    def __init__(self, root_directory):
+        self.root_directory = os.path.expanduser(root_directory)
+        self.error_files = {}
+        self.video_extensions = ('.mp4', '.avi', '.mov', '.mkv', '.wmv')
+        self.image_extensions = ('.jpg', '.jpeg', '.png', '.gif', 'webp')
+
+    def rename_files(self, file_type='all', dry_run=False):
+        current_date = datetime.datetime.now().strftime("%m%d")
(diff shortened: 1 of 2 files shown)


grzegorzkrukowski (changes requested):
Hey ! 

Thanks for your contribution !

I left some comments on the supported files, but my biggest concern is the hardcoded limitation to 2 digits.
Do you think it makes sense to calculate it dynamically based on the amount of files to be renamed?
You are already collecting them into an array, so you could maybe do something like:
```
total_files = len(files_to_rename)
num_digits = max(1, math.ceil(math.log10(total_files + 1)))
```

Or make it a max(2, ) if you think 2 digits makes more sense even for <10 files.

Oh and there is a typo in the filename - "vedios" when it should be "videos" ? also in PR name.

StepaniaH:
> Hey !
> 
> Thanks for your contribution !
> 
> I left some comments on the supported files, but my biggest concern is the hardcoded limitation to 2 digits. Do you think it makes sense to calculate it dynamically based on the amount of files to be renamed? You are already collecting them into an array, so you could maybe do something like:
> 
> ```
> total_files = len(files_to_rename)
> num_digits = max(1, math.ceil(math.log10(total_files + 1)))
> ```
> 
> Or make it a max(2, ) if you think 2 digits makes more sense even for <10 files.
> 
> Oh and there is a typo in the filename - "vedios" when it should be "videos" ? also in PR name.

Hey, expert, thanks for your review!

In my new commit, I try to fix the following points:
1. Add some common formats of video and picture.
2. Calculate the number dynamically based on the amount of files to be
renamed.
3. Fix the bugs of optional parameters.
4. The typos.

Thanks again.

grzegorzkrukowski:
> > Hey !
> > Thanks for your contribution !
> > I left some comments on the supported files, but my biggest concern is the hardcoded limitation to 2 digits. Do you think it makes sense to calculate it dynamically based on the amount of files to be renamed? You are already collecting them into an array, so you could maybe do something like:
> > ```
> > total_files = len(files_to_rename)
> > num_digits = max(1, math.ceil(math.log10(total_files + 1)))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Or make it a max(2, ) if you think 2 digits makes more sense even for <10 files.
> > Oh and there is a typo in the filename - "vedios" when it should be "videos" ? also in PR name.
> 
> Hey, expert, thanks for your review!
> 
> In my new commit, I try to fix the following points:
> 
> 1. Add some common formats of video and picture.
> 2. Calculate the number dynamically based on the amount of files to be
>    renamed.
> 3. Fix the bugs of optional parameters.
> 4. The typos.
> 
> Thanks again.

All right ! great stuff - thanks

PR #982, opened 2025-02-20
feat: add script to open clipboard URLs in Chrome

iloveitaly:
## Description

copy a list of urls, open them all in chrome

## Type of change

- [x] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Screenshot

NA

NA

## Checklist

### DIFF
--- a/commands/browsing/open-clipboard-urls-in-chrome.sh
+++ b/commands/browsing/open-clipboard-urls-in-chrome.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Open URLs in Clipboard in Chrome
+# @raycast.mode silent
+# @raycast.packageName Browsing
+
+# Optional parameters:
+# @raycast.icon 🔗
+
+# Documentation:
+# @raycast.author Michael Bianco
+# @raycast.authorURL https://github.com/iloveitaly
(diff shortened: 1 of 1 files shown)


dehesa:
Thank you for the contribution @iloveitaly, but since the introduction of quicklinks, we encourage people to use that instead of small script files.

iloveitaly:
Looks like quick links cannot run a script though, correct? What's the best way to implement something like this script using a quick link?

unnamedd:
@iloveitaly, that's correct. @dehesa's point here is that even though you won't be able to run a Script Command, Quicklinks allow you to specify where you want to open anything you want. That's why the Raycast team decided to incentivize the use of Quicklinks more instead of accepting new Script Commands that do what Quicklinks already allow the user to do. Also, you can share/export Quicklinks if you want to share them with someone.

iloveitaly:
@unnamedd Got it, so the solution here is to have a quick link open a script in the terminal?

unnamedd:
Actually, using Quicklinks, you won't need that because, with Quicklinks, you can define where you will open your things. It is much simpler than just writing a Script Command for that.

But yes, if you want, you can point to a local file in the link field (inside the "Create a Quicklink" form) and make it use the Terminal to execute it. I've never tried it, though, but I assume it would work nicely.

PR #996, opened 2025-07-01
Added Reboot and Log out scripts

VadimOza:
## Description

From time to time I need to reboot or logout of the mac. It would be very convenient to do that directly from Raycast instead of clicking this by mouse

## Type of change

- [X] New script command

## Screenshot

Regular reboot/log out by mac os

Only for MacOS

## Checklist

### DIFF
--- a/commands/system/logout.applescript
+++ b/commands/system/logout.applescript
@@ -1,0 +1,18 @@
+#!/usr/bin/osascript
+
+# Required parameters:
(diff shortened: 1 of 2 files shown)


dehesa:
Hi @VadimOza, thank you for the contribution, but we already have this natively in Raycast. Just type `Restart` or `Log out`. Don't you see it?

PR #1002, opened 2025-09-07
Add script to send Mermaid Diagram to Kroki Diagram Generator

zmx:
## Description

Send Mermaid diagram to Kroki Diagram Generator https://kroki.io/

## Type of change

- [V] New script command

## Screenshot

It copy Mermaid code in clipboard and send to https://kroki.io/

image

No, https://kroki.io/ is a public API endpoint

## Checklist

- [V] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)

### DIFF
--- a/commands/conversions/kroki-mermaid.sh
+++ b/commands/conversions/kroki-mermaid.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title My First Script
+# @raycast.mode sildent
+# @raycast.packageName Raycast Scripts
+#
+# Optional parameters:
+# @raycast.icon 🤖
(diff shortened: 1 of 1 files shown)


grzegorzkrukowski (changes requested):
Hey - thank you for your contribution

The script doesn't work as it is:

1. Type in mode:
`# @raycast.mode sildent`
should be:
`# @raycast.mode silent`

2. When copying samples from Kroki to PasteBoard, I am getting an error:
`No diagram type detected matching given configuration for text`
Probably something with encoding?

3. User experience
It would be a better user experience to check if user has python3 installed. If not thowing a readable error so users can figure out what they need to do to make this script work.
Also checking if pbpaste is empty or not before sending it to the kroki.io would be nice to have.

PR #1003, opened 2025-09-24
adding magic keyboard switcher script

blastik:


grzegorzkrukowski (approved):
Thank you for your contribution

PR #1004, opened 2025-09-27
Added SSH Tunnel SOCKS Proxy script command

AndriiBarabash:
## Description

Added a script command that creates an SSH tunnel and then turns on the SOCKS proxy through this tunnel.

## Type of change

- [x] New script command

## Screenshot

Screenshot 2025-09-27 at 20 45 39

It's a template, so it requires the variables to be set. Additionally, since it's not an interactive script, the SSH connection must be established without requiring password input, using SSH keys.

## Checklist


grzegorzkrukowski (approved):
Thank you for your contribution

PR #1007, opened 2025-10-24
Added delay to paste

AlexGadd:
If keys are still held down while it starts typing this can trigger commands or alternative characters to be typed, may need to add a description to warn people if they don't understand how this script works? Not sure if that's standard practice for these commits

## Type of change

If keys are still being held down it can cause unexpected outcomes so I added a delay before it starts "typing"

- [ ] New script command
- [ ] Bug fix
- [ ] Documentation update
- [ ] Toolkit change
- [ ] Other (Specify)

## Checklist

### DIFF
--- a/commands/system/paste-clipboard.sh
+++ b/commands/system/paste-clipboard.sh
@@ -14,8 +14,30 @@
 # @raycast.author AlexGadd
 # @raycast.authorURL https://raycast.com/AlexGadd
 
+sleep 0.3
+#returns "true" or "false" if key is held down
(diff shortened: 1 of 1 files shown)


AlexGadd:
Yes, however, part of the issue is that it runs immediately after the key combination is pressed, so you’d have to have inhuman speed to avoid any letters that are typed being effected. I could add a delay plus detecting if keys are down and showing an error window telling them to release the keys. I’ll try add that this week

nhojb (approved):
Looks good.

I guess you could use a single javascript script that returns a simple true/false if _any_ of the modifiers are down? But I guess if the overhead of running 5 scripts is not noticeable then it doesn't matter too much.

Happy to merge as-is.

PR #1030, opened 2026-03-31
Update safari-create-reading-list-item.applescript

JNBARY:
## Description

Modified the script to optionally use the URL of the webpage currently open in Safari as Link for the new Reading List item.

## Type of change

- [X] Improvement of an existing script

## Screenshot

By using "open" for the "Link"-argument the webpage currently open in Safari will be added to the Reading List.

No additional steps needed during installation.

## Checklist

### DIFF
--- a/commands/apps/safari/safari-create-reading-list-item.applescript
+++ b/commands/apps/safari/safari-create-reading-list-item.applescript
@@ -8,7 +8,7 @@
 # Optional parameters:
 # @raycast.packageName Safari
 # @raycast.icon images/safari.png
-# @raycast.argument1 { "type": "text", "placeholder": "Link, use open for current page" }
+# @raycast.argument1 { "type": "text", "placeholder": "Link, use 'open' for current Safari page" }
 # @raycast.argument2 { "type": "text", "placeholder": "Title", "optional": true }
 
 # @Documentation:
--- a/commands/apps/safari/safari-create-reading-list-item.applescript
+++ b/commands/apps/safari/safari-create-reading-list-item.applescript
@@ -13,7 +13,10 @@
 
 # @Documentation:
 # @raycast.author Thomas Paul Mann
+# @raycast.author Thomas Paul Mann
 # @raycast.authorURL https://github.com/thomaspaulmann
+# @raycast.author JNBARY
+# @raycast.authorURL https://github.com/JNBARY
 # @raycast.description Add a new Reading List item with the given URL. Allows a custom title to be specified.
 
 on run argv
--- a/commands/apps/safari/safari-create-reading-list-item.applescript
+++ b/commands/apps/safari/safari-create-reading-list-item.applescript
@@ -8,7 +8,7 @@
 # Optional parameters:
 # @raycast.packageName Safari
 # @raycast.icon images/safari.png
-# @raycast.argument1 { "type": "text", "placeholder": "Link" }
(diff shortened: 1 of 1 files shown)


commands/apps/safari/safari-create-reading-list-item.applescript
```diff
@@ -8,7 +8,7 @@
 # Optional parameters:
 # @raycast.packageName Safari
 # @raycast.icon images/safari.png
-# @raycast.argument1 { "type": "text", "placeholder": "Link" }
+# @raycast.argument1 { "type": "text", "placeholder": "Link, use open for current page" }
```
unnamedd:
```suggestion
# @raycast.argument1 { "type": "text", "placeholder": "Link, use 'open' for current Safari page" }
```

unnamedd (changes requested):
Hi @JNBARY, thanks for the contribution to the repository.

There's only one thing that, in my opinion, you need to improve: the placeholder message that will be shown in the root search. Even though the Script Command here is a Safari one, it is important to add the word "Safari" to avoid confusion. Additionally, you should clarify which word will be used in your action.

However, on second thought, I believe it would be much better if, instead of requiring the user to type "open" (which is essentially giving them an extra task) to add the page to the Reading List, you might want to discard this PR and create a separate Script Command exclusively for that purpose. The user would type into the root search: "Add Open Page to Reading List" and voilà, everything is set.

What do you think?

commands/apps/safari/safari-create-reading-list-item.applescript
JNBARY: Hi @unnamedd, your suggestion regarding the separate command sounds reasonable.
For my use case however, I like the simplicity of having one command do both and would thus commit the command with the changes you suggested. 

If there is a need for a second command that only ads the current Safari page, I'm open to someone else creating it.

Thank you for your work maintaining the repo!

commands/apps/safari/safari-create-reading-list-item.applescript
unnamedd, line 17: There's one last thing I forgot to bring here, it is highly recommended you append yourself as an author of the Script Command too, after all, we like to give and receive credits and Thomas would love to see you are contributing to his original Script Command.

```suggestion
# @raycast.author Thomas Paul Mann
# @raycast.authorURL https://github.com/thomaspaulmann
# @raycast.author JNBARY
# @raycast.authorURL https://github.com/JNBARY
```

unnamedd (changes requested):
Only one last thing here I forgot to ask you to add.

commands/apps/safari/safari-create-reading-list-item.applescript
unnamedd: For me, it is all good. As you are not changing the main behavior of the Script Command, it's fine. If you were changing it a lot, then it would be most unlikely to be accepted, but that is not the case.

commands/apps/safari/safari-create-reading-list-item.applescript
JNBARY: Thank you for the credit!
I would not have considered the addition that noteworthy, but of course will take it.

unnamedd (approved):
Nice! It is approved! 

As soon as our internal friends at Raycast can, they will merge it! :D

grzegorzkrukowski (approved):

PR #1032, opened 2026-04-14
Raycast Toolkit Improvements

unnamedd:
## Description

Finally, after a few years without giving any kind of love to our beloved `Toolkit`, here we are with a few improvements in the project.

In this pull request, I am introducing improvements and new support.

-  Upgrade in the external libraries
-  Support for Swift 6
-  Support for Swift async/await
-  New meta tag on Script Commands: `@raycast.platform`. This is due to the new support of Raycast for the Windows platform.
-  New language: PowerShell
-  Improvements in icon detection
-  Remove support for the original report
- Update the documentation in the main README.md
- Update the copyright in the file headers

## Type of change

- [ ] New script command
- [ ] Bug fix
- [ ] Documentation update
- [x] Toolkit change
- [ ] Other (Specify)

## Checklist


Tools/Toolkit/.editorconfig
```diff
@@ -0,0 +1,8 @@
+[*.swift]
+indent_style = space
+indent_size = 2
+tab_width = 2
+end_of_line = crlf
```
grzegorzkrukowski: we do have `crlf` here, but .swiftformat sets --linebreaks lf

Should we go with? 
```
[*.swift]                                                                                                                                                                                                                                                 
end_of_line = lf 
                                                                                                                                                                                                                                                            
[*.ps1]
end_of_line = crlf
```

Tools/Toolkit/.swift-version
```diff
@@ -0,0 +1 @@
+6.2
```
grzegorzkrukowski: We have 6.2 here, but 6.0 in swift-tools-version & swiftLanguageModes has .v6.m - should we stick to 6.0 for now?

unnamedd: Absolutely!

Tools/Toolkit/.editorconfig
unnamedd: Thanks for sharing this!

grzegorzkrukowski:
Hey @unnamedd - I've been looking into that again as I found few more things.

The generated `extensions.json` is broken - "paths" are truncated. 
If you try to generate it locally, you will find "system → stem, culture → lture, home → me".

I refactored a bit the URL handling, need to test things again next week to make sure there are no more regressions like that.

unnamedd:
Nice @grzegorzkrukowski! There's only one thing that puzzled me, and it was basically about the reason for reverting from "Awesome Script Commands" to the original "Raycast Script Commands" and also removing the link from the main README. Currently, there's no indication of it anywhere, that's why I decided to promote it in the main README. After all, everything in this repository belongs to Raycast, and I see no reason not to promote it in the main README.

Regarding the `extension.json`, I will have to replace the whole languages section in the root of the JSON because we already fixed the problem with `pwsh`, but it is not being replaced with the new `PowerShell`. Most likely, this is one of the problems we are facing with the generated JSON.

grzegorzkrukowski:
> Nice @grzegorzkrukowski! There's only one thing that puzzled me, and it was basically about the reason for reverting from "Awesome Script Commands" to the original "Raycast Script Commands" and also removing the link from the main README. Currently, there's no indication of it anywhere, that's why I decided to promote it in the main README. After all, everything in this repository belongs to Raycast, and I see no reason not to promote it in the main README.
> 
> Regarding the `extension.json`, I will have to replace the whole languages section in the root of the JSON because we already fixed the problem with `pwsh`, but it is not being replaced with the new `PowerShell`. Most likely, this is one of the problems we are facing with the generated JSON.

We can keep the link from the main Readme, but I am not 100% sure we want to change "Raycast Script Commands" as this is how we reference it in other places from Raycast + it will make discovery easier.

I will get an opinion from our legal team on the naming to check if we are ok with removing "Raycast" from it

unnamedd:
> We can keep the link from the main Readme, but I am not 100% sure we want to change "Raycast Script Commands" as this is how we reference it in other places from Raycast + it will make discovery easier.
> 
> I will get an opinion from our legal team on the naming to check if we are ok with removing "Raycast" from it

Nice, give a check on them. Regarding the name, it definitely doesn't not affect me, when I changed it there, was to having something broader in mind, but if it could cause legal implications, (as you) I prefer not to touch in it. 
Also, when you say it is referred in other places, do you mean: that list of Script Commands or the "Raycast Script Command" itself. Because then, we might be talking about two different things in here. 

Regarding the README, I do think we need to increase the discoverability of that index of Script Commands, especially now Raycast arrived on Windows and we are starting to receive PRs for the Windows platform. It will be useful and more discoverable.

grzegorzkrukowski:
> > We can keep the link from the main Readme, but I am not 100% sure we want to change "Raycast Script Commands" as this is how we reference it in other places from Raycast + it will make discovery easier.
> > I will get an opinion from our legal team on the naming to check if we are ok with removing "Raycast" from it
> 
> Nice, give a check on them. Regarding the name, it definitely doesn't not affect me, when I changed it there, was to having something broader in mind, but if it could cause legal implications, (as you) I prefer not to touch in it. Also, when you say it is referred in other places, do you mean: that list of Script Commands or the "Raycast Script Command" itself. Because then, we might be talking about two different things in here.
> 
> Regarding the README, I do think we need to increase the discoverability of that index of Script Commands, especially now Raycast arrived on Windows and we are starting to receive PRs for the Windows platform. It will be useful and more discoverable.

Thanks for understanding. I checked with Petr and we would like to keep the "Raycast Script Commands" - of course we can re-add the link from the main Reaedme.

unnamedd:
@grzegorzkrukowski no problem at all for me. I will re-add the link in the README and I guess we are good to go for a merge. What do you think?

grzegorzkrukowski:
Hey @unnamedd - is there anything else you want to modify in this PR in regard of the problems I found?

Especially the generation of the extensions.json problems revealed ? Or are we ok with the fixes I pushed?

From my perspective we can merge it whenever you want

grzegorzkrukowski (approved):

unnamedd:
@grzegorzkrukowski I was expecting for you haha

Let's merge it, and in case something goes wrong with the extensions.json generator, I can fix and submit another PR and ping you.

unnamedd:
keep it open.

unnamedd:
keep it open

grzegorzkrukowski:
Oh man, sorry for a huge delay. GitHub notifications are so hard to follow up and keep up with and with our V2 release I was completely overwhelmed and forgot about this PR :( Feel free to catch me on Slack and keep bugging me in the future if this getting delayed so much.

unnamedd:
hahaha don't worry @grzegorzkrukowski, I already appreciate a lot with time helping this humble PR here! Thanks for merging it!

PR #1037, opened 2026-05-27
Roman Numerals

kyellareddy:
## Type of change

- [x] New script command

## Screenshot

Video of it working:
Note that it also says out loud "XIV is equal to 14" at the same time that it shows in the toast.

https://github.com/user-attachments/assets/15e2cddb-0c07-45e3-8a5a-66d056ebb8d2

## Checklist

- [.] I have read [Contribution Guidelines](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md)


kyellareddy:
@dehesa

commands/productivity/Roman Numerals.py
```diff
@@ -0,0 +1,134 @@
...
+# @raycast.author Kailash Yellareddy
+# @raycast.authorURL https://github.com/kyellareddy
+
+# @raycast.argument1 { "type": "text", "placeholder": "Roman Numeral", "optional": false }
+
+import sys
+import subprocess
+
+roman_numeral = sys.argv[1] if len(sys.argv) > 1 else ""
+r = roman_numeral.upper()
```
grzegorzkrukowski: This seems to be never used and a line above as well

commands/productivity/Roman Numerals.py
```diff
@@ -0,0 +1,134 @@
...
+        return 50
+    if ch.upper() == 'C':
+        return 100
+    if ch.upper() == 'D':
+        return 500
+    if ch.upper() == 'M':
+        return 1000
+    return 0
+
+def romanToDecimal(s):
```
grzegorzkrukowski: This returns int or error string - probably returning None on error will be cleaner?

commands/productivity/Roman Numerals.py
```diff
@@ -0,0 +1,134 @@
...
+r = roman_numeral.upper()
+
+def speak(text):
+    subprocess.run(["say", text])
+
+def speak_roman(roman):
+    return " ".join(roman)
+
+def value(ch: str):
+    if ch.upper() == 'I':
```
grzegorzkrukowski: The .upper() call feel redundant here as called already does .upper() - or maybe caller should not be doing it?

grzegorzkrukowski (changes requested):
Hey @kyellareddy - thanks for your contribution and sorry for a delayed review.

There are several scripts added to this PR in different folders - is it intentional? 
It feels like other scripts should be added as a separate PRs as they are completely not related to "Roman Numberals" script itself.

As for the Roman Numerals itself that is the goal of this PR:
Every invocation runs speech as a `subprocess.run` blocking also the toast which is quite suprising effect. 
Did you consider making it maybe optional or at least non-blocking for script execution by using `subprocess.Popen` ?

The biggest problem is:
- `@raycast.packageName` is missing
- The filename convention is lowercase dash-case: `roman-numerals.py`

I also left some inline suggestions for the code of the Roman Numerals - but they are optional - once you fix the file naming and missing metadata and remove unnecessary files, we can merge it as it is.

kyellareddy (commented):
Delete Zip Code Lookup

kyellareddy:
First, that was not intentional. I accidentally added them all at once and didn't realize it until after I submitted the PR. I couldn't figure out how to delete them from the PR.

I completed all the changes you requested. However, I am still not sure how to delete the other scripts.
Is that fine, or should I just close this and open a new PR?

PR #1039, opened 2026-08-25
roman-numerals

kyellareddy:
## Description

Script command that takes any Roman numeral and outputs the equivalent in Arabic numerals (1, 2, 3, etc.).
## Type of change
- [x] New script command

## Screenshot
Video of it working:
Note that it also says out loud "XIV is equal to 14" at the same time that it shows in the toast.

https://github.com/user-attachments/assets/343aa6c5-f3f0-425b-964f-de1545e8897c

## Checklist

### DIFF
--- a/commands/productivity/Roman Numerals.py
+++ b/commands/productivity/Roman Numerals.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+
+# Required parameters:
+# @raycast.schemaVersion 1
+# @raycast.title Roman Numerals
+# @raycast.mode compact
+
+# Optional parameters:
+# @raycast.icon 🔢
+
+# Documentation:
+# @raycast.author Kailash Yellareddy
+# @raycast.authorURL https://github.com/kyellareddy
+
+# @raycast.argument1 { "type": "text", "placeholder": "Roman Numeral", "optional": false }
+
+import sys
+import subprocess
+
+
+def speak(text):
+    subprocess.run(["say", text])
+
+def speak_roman(roman):
+    return " ".join(roman)
+
+def value(ch: str):
+    if ch == "I":
+        return 1
+    if ch == "V":
+        return 5
+    if ch == "X":
+        return 10
+    if ch == "L":
+        return 50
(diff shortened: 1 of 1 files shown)


grzegorzkrukowski (changes requested):
Thanks for the contribution! 

There are a few things to fix before this can be merged:

## Repository conventions
- **File name:** please use lowercase dash-case, e.g. `roman-numerals.py` (see [File naming convention](https://github.com/raycast/script-commands/blob/master/CONTRIBUTING.md#file-naming-convention)).
- **Metadata:** please add `@raycast.packageName Conversions` (required in this repo) and a `@raycast.description`.
- **Location:** this is a converter, so it belongs in `commands/conversions/` next to scripts like `hex-to-rgb.sh` and `what-day-is.py`. The `productivity` folder is for tool-specific subfolders.

## Behavior
- **Invalid Roman numerals are accepted.** The parser only checks that each character is a valid symbol, not that the sequence is well-formed. For example, `IIII` → 4, `IC` → 99, `VX` → 5, `IIX` → 10, `LL` → 100, `MMMM` → 4000. One way to fix it is to convert the parsed number back with `decimalToRoman` and reject the input if the round trip does not reproduce the original string.
- **Numbers above 3999** currently produce a long string of `M`s (try `99999`) and then read every one of them aloud. Since the script already knows the limit is 3999, it should just reject larger input.
- **Speech is unconditional.** Every run shells out to `say` and blocks until it finishes, including on error paths. That is unexpected for a converter. I would either drop it, or make it opt-in with a dropdown argument (e.g. `@raycast.argument2 { "type": "dropdown", "placeholder": "Speak", "optional": true, "data": [...] }`).

## Minor
- `value()` can be a dict lookup.
- "That is not a valid English numeral" → "That is not a valid number".
- `romanToDecimal` returns either an `int` or an error string. Returning `None` on failure (or raising) would make the caller simpler.

Happy to take another look once these are addressed!

kyellareddy:
Done, I added a new commit.

